PR #2, opened 2014-07-24
Support any type that implements fmt.Stringer in ObjectIdHex and IsObjectIdHex instead of simple string

FeNoMeNa:
This change allows using of any type that implements fmt.Stringer in ObjectIdHex and IsObjectIdHex functions.

Example:

If we have type `type UserId string`, we have to typecast UserId to string.
With this change we simply pass the UserId object without typecast.


fatih:
I don't believe that would work. You need to implement the `String()` method for every type you want to pass to make use of the Stringer interface. That means now it's not possible to pass a simple string type anymore :) http://play.golang.org/p/myVcRtnc_A

I suggest to revisit this PR or even close it.

niemeyer:
@fatih is spot on.

PR #6, opened 2014-08-06
Add a very basic readme

vkarpov15:
Hi Gustavo,

Just wanted to add a sensible README as well as make sure I'm running the tests properly :)

Cheers,
Val


niemeyer:
Thanks again. If that's okay, I'd prefer to keep project details at the project page to avoid having to maintain multiple sources. I've added a trivial README.md, though, and also updated labix.org/mgo to include these testing instructions.

PR #10, opened 2014-08-18
MGO-35 add support for ServiceHostname for GSSAPI, test for ServiceName

vkarpov15:
See corresponding code in server's MongoClient [here](https://github.com/mongodb/mongo/blob/be05259d36ec5f44c821c8f2d103a4ee83ebaf53/src/mongo/client/sasl_client_session.cpp#L279-285)

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -246,7 +246,13 @@ func (socket *mongoSocket) loginPlain(cred Credential) error {
 }
 
 func (socket *mongoSocket) loginSASL(cred Credential) error {
-	sasl, err := saslNew(cred, socket.Server().Addr)
+	var sasl saslStepper;
+	var err error;
+	if len(cred.ServiceHostname) > 0 {
+		sasl, err = saslNew(cred, cred.ServiceHostname)
+	} else {
+		sasl, err = saslNew(cred, socket.Server().Addr)
+	}
 	if err != nil {
 		return err
 	}
--- a/auth.go
+++ b/auth.go
@@ -249,7 +249,7 @@ func (socket *mongoSocket) loginSASL(cred Credential) error {
 	var sasl saslStepper
 	var err error
 	if len(cred.ServiceHost) > 0 {
-		sasl, err = saslNew(cred, cred.ServiceHostname)
+		sasl, err = saslNew(cred, cred.ServiceHost)
 	} else {
 		sasl, err = saslNew(cred, socket.Server().Addr)
 	}
--- a/auth.go
+++ b/auth.go
@@ -246,9 +246,9 @@ func (socket *mongoSocket) loginPlain(cred Credential) error {
 }
 
 func (socket *mongoSocket) loginSASL(cred Credential) error {
-	var sasl saslStepper;
-	var err error;
-	if len(cred.ServiceHostname) > 0 {
+	var sasl saslStepper
+	var err error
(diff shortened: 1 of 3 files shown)


niemeyer:
That's great, thanks Valeri!

It looks generally good. I'll add a few short notes inline.

auth.go
```diff
@@ -246,7 +246,13 @@ func (socket *mongoSocket) loginPlain(cred Credential) error {
 }
 
 func (socket *mongoSocket) loginSASL(cred Credential) error {
-	sasl, err := saslNew(cred, socket.Server().Addr)
+	var sasl saslStepper;
+	var err error;
```
niemeyer: Please drop the commas at the end of the line, and also run go fmt to ensure the code is properly formatted according to the usual conventions. There are spacing issues below.

auth_test.go
```diff
@@ -936,3 +936,73 @@ func (s *S) TestAuthKerberosURL(c *C) {
...
+
+	c.Logf("Connecting to %s...", kerberosHost)
+        session, err := mgo.Dial(kerberosHost)
+        c.Assert(err, IsNil)
+        defer session.Close()
+
+	c.Logf("Authenticating with incorrect service name...")
+        err = session.Login(cred)
+        c.Assert(err, ErrorMatches,
+		".*Server wrong/mmscustmongo.10gen.me@10GEN.ME not found.*")
```
niemeyer: Please move it all in a single line here and in the equivalent line in the next test. Given the usual term size these days, we're not following the 80 columns rule.

session.go
```diff
@@ -303,6 +303,11 @@ type DialInfo struct {
 	// mechanism. Defaults to "mongodb".
 	Service string
 
+	// ServiceHostname defines which hostname to use when authenticating
+        // with the GSSAPI mechanism. If not specified, defaults to the MongoDB
+	// server's address.
+	ServiceHostname string
```
niemeyer: Let's please use the field name "ServiceHost", to follow usual Go conventions (the "host" term is abundant in Go code.. see the standard net package for example).

vkarpov15:
Comments addressed. Sorry about the whitespace issues, vim's copy/paste seems to maul tabs

niemeyer:
Please run tests. :-)

vkarpov15:
Yep good call

niemeyer:
Sorry for bothering, but spacing is still messed up. See lines 307 and 606.

I'll post a couple of git hooks to make sure we don't need to fight those trivial details on reviews.

vkarpov15:
Yeah that would be good. And I really need to get myself a real Linux machine instead of just SSH-ing into EC2 boxes :)

vkarpov15:
Hey @niemeyer any more comments re: this?

niemeyer:
No, this seems ready to go in.

vkarpov15:
Any idea when? I got a code review for mongo-tools blocking on this :)

niemeyer:
I suppose you need a new release to unblock that, rather than just a merge?
We can do either.

niemeyer:
Merged on v2-dev on f047e079.

vkarpov15:
I think merging should be sufficient for us, we just need the gitspec in the git repo so `gpm` can do its thing

niemeyer:
Cool, it's in.

PR #11, opened 2014-08-20
golint cleanup

ryancox:
This patch cleanups up 100% of the issues reported by golint. 

Note: this PR is **NOT** intended to be merged without discussion and thought as it contains breaking changes to public APIs. 

The main two classes of changes were:
- fooId -> fooID
- Undocumented public functions

I also had to crank up some sleeps in the test cases in order for them to pass; perhaps systems are slower.


niemeyer:
Yeah, thanks for getting us to know golint's preferences, but even without debating which of these should actually be done, we simply cannot do these breaking changes without a fantastic reason.

PR #12, opened 2014-08-25
Make it possible to run tests on windows (with a little cygwin help)

vkarpov15:
Hi Gustavo,

This PR makes it actually possible to run tests in the native windows shell:

image

Making supervisor run on native windows is a lost cause. With a little extra finagling you can make it behave reasonably on Cygwin though.

Comments eagerly awaited :)


vkarpov15:
Hi @niemeyer, any feedback on this? I'm going to put in a PR for Windows kerberos soon that depends on these changes.

niemeyer:
It looks mostly good, but there are a few points to address:

Why is Stop not supported on Windows?  This is just stopping and restarting the process via supervisord

Duplicating the whole configuration is not great. There must be a nice way to avoid it.

What's the deal with the DIRECTORY environment variable?

The very specific ``uname` == 'CYGWIN_NT-6.1'` check doesn't provide a lot of confidence that other people will be able to reproduce the results.

vkarpov15:
1) Supervisor doesn't work on native windows, and `cmd.Exec()` runs commands against the native windows shell, so supervisor will blow up.

2/3) The reasoning for the DIRECTORY variable and config duplication is the usual cygwin vs native mess: even if you run supervisor on cygwin, mongod still needs the native windows paths because the mongod windows executable isn't built with cygwin. However, when you run supervisor on cygwin, `%(here)s` gives you the cygwin path. In addition, mongod on windows doesn't support `--fork`.

I agree the configuration isn't very general, but I don't have any better ideas for getting the tests to run on MCI EC2 windows boxes without swapping out supervisor entirely.

niemeyer:
We can swap supervisord for something better and that works in Windows. Any suggestions? Any idea if runit works well on Windows?

vkarpov15:
Not terribly familiar with runit. I know the node driver has its own windows-specific cluster manager using the windows `taskkill` command. It wouldn't be too difficult to adapt such a codebase for Go (all you really need is a clean API for killing and starting a process, which should be easy enough with build tags), and would give us the added benefit of not having a dependency on python for testing.

However, I think that's outside the scope of this particular PR, since I just need the Kerberos tests and those don't require `make startdb`. If I just got rid of the changes to `setup.sh` and `testdb/supervisord-cygwin.conf` would this PR be mergeable?

niemeyer:
Yes, I'm definitely happy to merge something simpler that solves your current problem.

To get this merged, can you please revert the changes to Thaw/Freeze, and instead just isolate the real offending call (syscall.Kill) in a similar fashion?

vkarpov15:
Ok I removed the supervisor changes and isolated syscall.Kill as per your suggestion

suite_test.go
```diff
@@ -159,7 +164,7 @@ func (s *S) pid(host string) int {
 }
 
 func (s *S) Freeze(host string) {
-	err := syscall.Kill(s.pid(host), syscall.SIGSTOP)
+	err := s.Stop(host)
```
niemeyer: That's bogus. The Stop method does not do what SIGSTOP does.

syscall_test.go
```diff
@@ -0,0 +1,15 @@
+// +build linux darwin
+
+package mgo_test
+
+import (
+	"syscall"
+)
+
+func (s *S) Stop(host string) error {
```
niemeyer: How does that even compile, when there's a S.Stop method in suite_test.go!?

Let's please have a single function here:

```
func kill(pid int, sig Signal) (err error) {
        return syscall.Kill(pid, sig)
}
```

Please make sure it works!

suite_test.go
vkarpov15: Good call, I got lazy and didn't test. Done.

syscall_test.go
vkarpov15: Can't do that, signal and SIGCONT+SIGSTOP don't exist on windows either. See commit I'm about to push.

syscall_win_test.go
```diff
@@ -0,0 +1,11 @@
+// +build windows
```
niemeyer: Please drop this line, and rename this file to syscall_windows_test.go. The go tool understands that convention.

suite_test.go
```diff
@@ -137,6 +139,9 @@ func (s *S) TearDownTest(c *C) {
 
 func (s *S) Stop(host string) {
 	// Give a moment for slaves to sync and avoid getting rollback issues.
+	if runtime.GOOS == "windows" {
+		panic("Stop() currently unsupported on windows!")
```
niemeyer: Please define this function:

```
func panicOnWindows() {
        if runtime.GOOS == "windows" {
                panic("the test suite is not yet fully supported on Windows")
        }
}
```

and change all the panics introduced to call it.

niemeyer:
It's looking good otherwise.

syscall_win_test.go
vkarpov15: Done

suite_test.go
vkarpov15: Done

vkarpov15:
Ok see how this one strikes you. Also keep in mind that this is blocking my work on #20, because without this change my workflow is a mess. Not my best idea to put this as a separate PR.

niemeyer:
Sorry it's causing you some trouble. I do appreciate your effort, though. Much better to understand and polish this in isolation.

I'll get it merged in a moment.

niemeyer:
Merged into v2-dev as bfb5e400.

PR #13, opened 2014-08-25
POC of byte copying for mongodump

3rf:
NOTE: The code below is just a hacked together proof of concept, and not meant to be merged!

Hi Gustavo,

As promised, here is a proof of concept for copying bson bytes directly from the server's response to a byte slice. First, let me give you the full context: 

For the mongodump tool, we need to be able to read all of the documents from a given collection and write them to disk, in raw BSON binary, as fast as possible. This specific (and relatively niche) use-case does not require unmarshalling the server's response--we, ideally, can just check for errors and then write the raw binary straight to disk. 

Our current prototype of mongodump does this by unmarshalling to a bson.Raw, which provides the usability we need, but is much too slow. The current prototype takes ~8 times longer to dump a collection than the existing C++ implementation. After profiling, it became apparent that this was caused by two things: (1) the large amount of reflect magic in bson Unmarshal and (2) garbage collection dealing with the allocation of millions of temporary buffers. 

The quick hack in this pull request gets around these issues by skipping all of the reflection in Unmarshal and by allowing us to reuse the same 16MB buffers, saving on garbage collection. Using this new NextCopyBytes method, our prototype runs at equivalent speed to the original C++ mongodump! 

Obviously, the code here is just an illustration of what we need. I am very interested in hearing your thoughts on this problem, and how you feel a solution could be best added to the current mgo API. Let me know of any questions you have. Happy to help with implementation any way I can.

Thanks!
-Kyle

### DIFF
--- a/session.go
+++ b/session.go
@@ -2794,6 +2794,94 @@ func (iter *Iter) Next(result interface{}) bool {
 	panic("unreachable")
 }
 
+// NextCopyBytes is a hack of Next(). It takes in an existing byte buffer and copies
+// the returned bson bytes from the db into it. It returns a boolean similar to Next()
+// and integer representing the number of bytes copied into the buffer. 
+func (iter *Iter) NextCopyBytes(out []byte) (bool, int) {
+	iter.m.Lock()
+	iter.timedout = false
+	timeout := time.Time{}
+	for iter.err == nil && iter.docData.Len() == 0 && (iter.docsToReceive > 0 || iter.op.cursorId != 0) {
+		if iter.docsToReceive == 0 {
+			if iter.timeout >= 0 {
+				if timeout.IsZero() {
+					timeout = time.Now().Add(iter.timeout)
+				}
+				if time.Now().After(timeout) {
+					iter.timedout = true
+					iter.m.Unlock()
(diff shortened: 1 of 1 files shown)


niemeyer:
Per mailing thread:

Kyle, I have reviewed the proof of concept over the weekend, and apparently it's just mimicking the behavior of Iter.Next, but copying the byte slice out of the internal buffer.

If I'm right about that, a couple of questions:
1. Have you tried to use a bson.Raw? If so, what was the impact? If you use one for the whole document, it's very close to being a raw copy of the byte slice, and we can definitely improve the performance further by fine tuning that code path. Which leads to the second point:
2. Do you have a small self-contained benchmark we could share, preferably one that might be part of the test suite permanently? I believe I might easily improve the performance of the important code path to be satisfactory, but I must be able to tell whether we've reached satisfaction or not, and hopefully we can keep that in place so it remains the case in the future.

niemeyer:
As discussed over email, commit 00004f72 turns unmarshaling into a bson.Raw into a trivial operation (~150ns locally), so we should be able to use the standard Next method and achieve the same benefits of the described proof of concept.

PR #16, opened 2014-09-02
Text search

ghost:
Adding text indexes and score sort.

There was a comment on Sort implying the following syntax:

``` go
query4 := collection.Find(nil).Sort("score:{$meta:textScore}")
```

but I think it makes more sense to be coherent with the index syntax and do:

``` go
query4 := collection.Find(bson.M{"$text": bson.M{"$search": "some text"}})
                    .Select(bson.M{"score": bson.M{"$meta": "textScore"}})
                    .Sort("$score:textScore")
```

Does that make sense?

It might be missing documentation. But I want to see if this might get accepted.

### DIFF
--- a/session.go
+++ b/session.go
@@ -910,14 +910,17 @@ func (db *Database) RemoveUser(user string) error {
 }
 
 type indexSpec struct {
-	Name, NS       string
-	Key            bson.D
-	Unique         bool ",omitempty"
-	DropDups       bool "dropDups,omitempty"
-	Background     bool ",omitempty"
-	Sparse         bool ",omitempty"
-	Bits, Min, Max int  ",omitempty"
-	ExpireAfter    int  "expireAfterSeconds,omitempty"
+	Name, NS         string
+	Key              bson.D
+	Unique           bool   ",omitempty"
+	DropDups         bool   "dropDups,omitempty"
+	Background       bool   ",omitempty"
+	Sparse           bool   ",omitempty"
+	Bits, Min, Max   int    ",omitempty"
+	ExpireAfter      int    "expireAfterSeconds,omitempty"
+	Weights          bson.D ",omitempty"
+	DefaultLanguage  string "default_language,omitempty"
+	LanguageOverride string "language_override,omitempty"
 }
 
 type Index struct {
@@ -932,14 +935,26 @@ type Index struct {
 	Name string // Index name, computed by EnsureIndex
 
 	Bits, Min, Max int // Properties for spatial indexes
+
+	// Properties for text indexes
+	DefaultLanguage  string
+	LanguageOverride string
 }
 
-func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
+type indexKeyInfo struct {
+	name    string
+	key     bson.D
+	weights bson.D
+}
+
+func parseIndexKey(key []string) (*indexKeyInfo, error) {
+	var keyInfo indexKeyInfo
+	isText := false
 	var order interface{}
 	for _, field := range key {
 		raw := field
-		if name != "" {
-			name += "_"
+		if keyInfo.name != "" {
+			keyInfo.name += "_"
 		}
 		var kind string
 		if field != "" {
@@ -947,7 +962,7 @@ func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
 				if c := strings.Index(field, ":"); c > 1 && c < len(field)-1 {
 					kind = field[1:c]
 					field = field[c+1:]
-					name += field + "_" + kind
+					keyInfo.name += field + "_" + kind
 				}
 			}
 			switch field[0] {
@@ -960,32 +975,40 @@ func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
 				// The shell used to render this field as key_ instead of key_2d,
 				// and mgo followed suit. This has been fixed in recent server
 				// releases, and mgo followed as well.
-				name += field + "_2d"
+				keyInfo.name += field + "_2d"
 			case '-':
 				order = -1
 				field = field[1:]
-				name += field + "_-1"
+				keyInfo.name += field + "_-1"
 			case '+':
 				field = field[1:]
 				fallthrough
 			default:
 				if kind == "" {
 					order = 1
-					name += field + "_1"
+					keyInfo.name += field + "_1"
 				} else {
 					order = kind
 				}
 			}
 		}
 		if field == "" || kind != "" && order != kind {
-			return "", nil, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
+			return nil, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
 		}
-		realKey = append(realKey, bson.DocElem{field, order})
+		if kind == "text" {
+			if !isText {
+				keyInfo.key = append(keyInfo.key, bson.DocElem{"_fts", "text"}, bson.DocElem{"_ftsx", 1})
+				isText = true
+			}
+			keyInfo.weights = append(keyInfo.weights, bson.DocElem{field, 1})
+		} else {
+			keyInfo.key = append(keyInfo.key, bson.DocElem{field, order})
+		}
 	}
-	if name == "" {
-		return "", nil, errors.New("invalid index key: no fields provided")
+	if keyInfo.name == "" {
+		return nil, errors.New("invalid index key: no fields provided")
 	}
-	return
+	return &keyInfo, nil
 }
 
 // EnsureIndexKey ensures an index with the given key exists, creating it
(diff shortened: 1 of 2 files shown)


niemeyer:
The original plan was to have something simpler as you suggest, and then came the idea of using a more flexible syntax to prevent potential future compatibility issues. That said, since I wrote this my feeling has changed back that this is not necessary, and a simpler model conforming to the current convention should work for a long time.

That said, there's a bug in the current proposal. The special metadata name is "textScore", not "score". The latter is the field name.

ghost:
Then do I understand correctly that you prefer the following syntax?

``` go
query4 := collection.Find(nil).Sort("score:{$meta:textScore}")
```

So the select will be automagically created from the Sort function, isn't it?
I can give a try to this idea and pull-req again.

And yes, you are right that my code right now is wrong.

niemeyer:
As I mentioned, a simpler model conforming to the current convention should work.

Something along the lines of:

```
$textScore:score
```

ghost:
I miss understood your comment. I fixed the issue. Does it look ok now? Anything to improve?

niemeyer:
There are several changes unrelated to the conversation above in the current pull request. It'd be good to trim it down to adding just that alone, and then adding a test. Other changes, if you still want to push them, may be proposed in a separate request.

ghost:
There is two commits that add two features:
- text indexes
- score sort

Can you point me to other unrelated changes?

I'll add some tests.

ghost:
I just added a test.

Does it look ok now? Does my implementation of the text indexes makes sense?

niemeyer:
Thanks. I'll provide a few comments inline.

session.go
```diff
@@ -910,14 +910,17 @@ func (db *Database) RemoveUser(user string) error {
...
+	Key               bson.D
+	Unique            bool   ",omitempty"
+	DropDups          bool   "dropDups,omitempty"
+	Background        bool   ",omitempty"
+	Sparse            bool   ",omitempty"
+	Bits, Min, Max    int    ",omitempty"
+	ExpireAfter       int    "expireAfterSeconds,omitempty"
+	Weights           bson.M ",omitempty"
+	Default_language  string ",omitempty"
+	Language_override string ",omitempty"
```
niemeyer: Please use standard Go names for the fields (e.g. DefaultLanguage) and use the field tag to adjust the name as necessary (see ExpireAfter above).

session.go
```diff
@@ -932,9 +935,19 @@ type Index struct {
...
+	Weights          map[string]int
+	DefaultLanguage  string
+	LanguageOverride string
 }
 
-func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
+func parseIndexKey(key []string, weights map[string]int) (name string, realKey bson.D, realWeights bson.M, err error) {
+	isText := false
+	realWeights = bson.M{}
+	for k, v := range weights { //FIXME: this is uggly
```
niemeyer: The whole logic around weight is not looking great. Too many result arguments, duplication, etc.

Let's implement this as part of the string itself:

```
$text,weight=1:foo
```

That said, let's please do that in a separate pull request. Just elide all references to weight here, and let's get this merged sooner rather than later.

session_test.go
```diff
@@ -2073,6 +2073,56 @@ func (s *S) TestSortWithBadArgs(c *C) {
...
 }
 
+func (s *S) TestSortScoreText(c *C) {
+	session, err := mgo.Dial("localhost:40001")
+	c.Assert(err, IsNil)
+	defer session.Close()
+
+	coll := session.DB("mydb").C("mycoll")
+
+	coll.EnsureIndex(mgo.Index{
```
niemeyer: Every function that returns an error should have the error checked.

session_test.go
```diff
@@ -2073,6 +2073,56 @@ func (s *S) TestSortWithBadArgs(c *C) {
...
+
+	query := coll.Find(M{"$text": M{"$search": "MongoDB"}})
+	query.Select(M{"score": M{"$meta": "textScore"}})
+	query.Sort("$textScore:score")
+	iter := query.Iter()
+
+	l := make([]string, 6)
+	r := struct{ A, B string }{}
+	for i := 0; i != len(l); i += 2 {
+		ok := iter.Next(&r)
```
niemeyer: Rather than guessing the precise response, please exhaust the iterator as usual (`for iter.Next(&r) { ... }`) appending r.A and r.B to a slice that is uninitialized to begin with (`var results []string` and `results = append(results, r.A, r.B)`), and then compare the outcome to the expected value as you're doing below.

session_test.go
```diff
@@ -2073,6 +2073,56 @@ func (s *S) TestSortWithBadArgs(c *C) {
...
+		l[i+1] = r.B
+	}
+
+	c.Assert(l, DeepEquals, []string{
+		"many",
+		"MongoDB MongoDB MongoDB MongoDB",
+		"MongoDB Manual",
+		"is an open-source document-oriented database",
+		"new features",
+		"MongoDB 2.6 is now available",
```
niemeyer: That's every single test value in the test set, and it's also hard to see why they're expected in that order at a glance.

Please:
1. Use "foo" as the matching string.
2. Use prefixes that are indicative of the test (`"just once: foo"`, `"many times: foo foo foo"`, etc).
3. Add sample values that match only on A and only on B (`"none"`)
4. Add a sample value that does not match at all.
5. Include in the sample value that does not match a third field `"c"` with `"ignore: foo"`.

The hint is that it should be clear at a glance why the order is as suggested.

niemeyer:
That's it for now. Thanks again for pushing this forward.

session.go
ghost: I can not remove all support for weights from this pull-request, the weights are needed for the text indexes. The text keys in the index are not in the list of keys, only in the list of weights. I can skip implementing the user input for weights, but I still need to put the weights = 1 in the index.

It's true that there is too many result arguments. What about returning (indexSpec, error) instead of name, key & weights?

ghost:
I think I addressed all your comments.

I can not remove all the weight support, it's needed by the index. I use indexSpec as return value of parseIndexKey to reduce the amount of return values. Is this ok?

session_test.go
```diff
@@ -2073,6 +2073,58 @@ func (s *S) TestSortWithBadArgs(c *C) {
...
+		"c": "ignore: foo",
+	})
+	c.Assert(err, IsNil)
+
+	query := coll.Find(M{"$text": M{"$search": "foo"}})
+	query.Select(M{"score": M{"$meta": "textScore"}})
+	query.Sort("$textScore:score")
+	iter := query.Iter()
+
+	var r struct{ A, B, C string }
```
niemeyer: Please drop C from the list.

session.go
```diff
@@ -932,9 +935,17 @@ type Index struct {
...
+	// Properties for text indexes
+	DefaultLanguage  string
+	LanguageOverride string
 }
 
-func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
+func parseIndexKey(key []string) (spec indexSpec, err error) {
+	name := ""
+	realKey := bson.D{}
+	weights := bson.M{}
```
niemeyer: Okay, I see why you want to return weights. Please return (*indexKeyInfo, error) with:

```
type indexKeyInfo struct {
    name string
    key bson.D
    weights bson.D
}
```

session.go
```diff
@@ -978,13 +989,25 @@ func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
 			}
 		}
 		if field == "" || kind != "" && order != kind {
-			return "", nil, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
+			return spec, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
+		}
+		if kind == "text" {
+			isText = true
+			weights[field] = 1
```
niemeyer: keyInfo.weights = append(keyInfo.weights, bson.DocElem{field, 1})

ghost:
Another iteration :)

session.go
```diff
@@ -932,14 +935,26 @@ type Index struct {
...
+}
+
+type indexKeyInfo struct {
+	name    string
+	key     bson.D
+	weights bson.D
 }
 
-func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
+func parseIndexKey(key []string) (keyInfo *indexKeyInfo, err error) {
```
niemeyer: Please drop the result names. They are implicit from the types.

session.go
```diff
@@ -932,14 +935,26 @@ type Index struct {
...
+	keyInfo = &indexKeyInfo{"", bson.D{}, bson.D{}}
```
niemeyer: var keyInfo indexKeyInfo

session.go
```diff
@@ -960,30 +975,38 @@ func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
...
 		}
-		realKey = append(realKey, bson.DocElem{field, order})
 	}
-	if name == "" {
-		return "", nil, errors.New("invalid index key: no fields provided")
+	if keyInfo.name == "" {
+		return nil, errors.New("invalid index key: no fields provided")
+	}
+	if isText {
+		keyInfo.key = append(keyInfo.key, bson.DocElem{"_fts", "text"}, bson.DocElem{"_ftsx", 1})
```
niemeyer: This does not match what happens in the shell. For example:

http://paste.ubuntu.com/8292084/

The text index key must be appended when it is seen, together with every other entry above. It can use such a boolean to track whether it has already appended or not, though.

session.go
```diff
@@ -960,30 +975,38 @@ func parseIndexKey(key []string) (name string, realKey bson.D, err error) {
...
 				} else {
 					order = kind
 				}
 			}
 		}
 		if field == "" || kind != "" && order != kind {
-			return "", nil, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
+			return nil, fmt.Errorf(`invalid index key: want "[$<kind>:][-]<field name>", got %q`, raw)
+		}
+		if kind == "text" {
```
niemeyer, line 998: The special string is "textScore" in Sort and in the MongoDB server. We can as well use it here.

niemeyer:
Okay, this is looking good. I suspect the comments above are the last iteration.

Thanks for keeping it up.

session.go
ghost: mgo's code don't use 'ensureIndex' in the collection, instead of that it insert the index directly on 'system.indexes'. After several tries I found out that this is the syntax expected by the system.indexes collection.

If you look at the example you just pasted, in the index itself the 'key' list don't include the text keys only 'weights' include them.

session.go
ghost, line 998: MongoDB uses textScore for the sort score, but for the indexes it uses 'text'. I think it's more clear to use the names that MongoDB uses:
- index -> text
- sort -> textScore

But if you prefer it the other way around, I'm ok with it and I'll change it.

niemeyer:
Any news on this?

ghost:
I was waiting for you to answer to my comments on your comments.

session.go
niemeyer: That wasn't the point. Note where "bar" ends in the text key. Your logic is always appending the text index at the end, but that's not what the shell does. Maybe it doesn't matter, but there's no reason to depart from that either.

ghost:
Updated, now I understood your point on the text indexes, and I hope I fixed it.

niemeyer:
This was merged in v2-dev with b3fc1f86a and will be in the next release.

Thanks for keeping it up!

ghost:
Great!!
Thank you for keeping with it.

PR #18, opened 2014-09-03
Fix #17; Slices conforming to bson.Setter are not delegated to for subdoc decoding

inlined:
Fix for https://github.com/go-mgo/mgo/issues/17


niemeyer:
That's not quite right. The logic in the `default` cause includes the handling of some slices.

I'm still looking at the use case, but the fix is not so straightforward if we do it.

PR #19, opened 2014-09-03
simple benchmarks for raw optimization 

3rf:
Added two benchmarks for reading into bson.Raw. The one we really need optimized is the "BenchmarkFindIterRaw" case, but I also included "BenchmarkUnmarshalRaw" because it might be nice to have something of a smaller scope too. Let me know if you have any questions.

-Kyle

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -1448,14 +1448,14 @@ type BenchRawT struct {
 	D []float64
 }
 
-func BenchmarkUnmarhsalStruct(b *testing.B) {
+func (s *S) BenchmarkUnmarhsalStruct(c *C) {
 	v := BenchT{A: "A", D: "D", E: "E"}
 	data, err := bson.Marshal(&v)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


niemeyer:
Brilliant, thank you!

Would be nice to have the benchmarks integrated in gocheck as well. The support is exactly equivalent. They just need to be made part of the suite, and the traditional (c *C) argument used instead.

3rf:
Ported the tests to gocheck. One potential issue: when running the original BenchmarkFindIterRaw benchmark, on my laptop I get
 `BenchmarkFindIterRaw        2000      99252834 ns/op`
but after porting the code to gocheck I get
`PASS: session_test.go:3261: S.BenchmarkFindIterRaw        200     559879123 ns/op`
which is a whole order of magnitude longer ns/op. Any idea what could be causing this? The benchmarks in the bson package didn't increase in time at all

niemeyer:
This was merged in v2-dev in 10662809. I'll follow up on the comments via email so we can keep the conversation in a single place.

PR #20, opened 2014-09-04
POC for Kerberos on Windows with SSPI

vkarpov15:
I added 5 new files:
- `kerberos_sspi.*`: these are thin wrappers around windows system calls that help with loading DLLs, they're copied exactly from the [node driver's SSPI implementation](https://github.com/christkv/kerberos/tree/master/lib/win32)
- `sasl_sspi.c`, `sasl_sspi.h`: some glue to make it easier to use the windows system calls from Go. Mostly, its just a C adaptation of [Eric Milkie's implementation for the mongo shell](https://github.com/mongodb/mongo/blob/b823d270b570cf4da5629959ee6bfc7cdfd69413/src/mongo/client/sasl_sspi.cpp)
- `sasl_sspi.go`: exposes the same API as `sasl.go` but uses Windows SSPI instead of SASL.

Still a couple outstanding issues, most notably that 1) I haven't added tests yet (see #12, among other limitations which we can discuss), and 2) only ANSI support for now, no unicode, because for some reason my testing setup is blowing up when I try to use unicode.


sasl/sasl_sspi.go
```diff
@@ -0,0 +1,144 @@
...
+	return cstr
+}
+
+func (ss *saslSession) Close() {
+	for _, cstr := range ss.stringsToFree {
+		C.free(unsafe.Pointer(cstr))
+	}
+
+	// Make sure we've cleaned up all the buffers we malloced when we're sure we don't need em anymore
+	if ss.authComplete || ss.errored {
```
niemeyer, line 101: If this is being closed, why would buffers not be freed? Close should mean "stop doing everything you are doing and free all resources".

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
+#include "kerberos_sspi.h"
```
niemeyer, line 2: These files need a header. Where did you get the code from? What's the copyright and the license?

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // Call the function
+  return (*pfn_deleteSecurityContext)(phContext);
+}
+
+/**
+ * Decrypt Message
+ */
+SECURITY_STATUS SEC_ENTRY _sspi_DecryptMessage(PCtxtHandle phContext, PSecBufferDesc pMessage, unsigned long MessageSeqNo, unsigned long pfQOP) {
```
niemeyer: Where is this being used?

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  decryptMessage_fn pfn_decryptMessage = NULL;
+
+  // Return error if library not loaded
+  if(_sspi_security_dll == NULL) return -1;
+  // Map function
+  pfn_decryptMessage = (decryptMessage_fn)GetProcAddress(_sspi_security_dll, "DecryptMessage");
+
+  // Check if the we managed to map function pointer
+  if(!pfn_decryptMessage) {
+    printf("GetProcAddress failed.\n");
```
niemeyer: That's not a good way to convey an error. GetProcAddr of what, from which library, and definitely not to stdout.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  deleteSecurityContext_fn pfn_deleteSecurityContext = NULL;
+
+  // Return error if library not loaded
+  if(_sspi_security_dll == NULL) return -1;
+  // Map function
+  pfn_deleteSecurityContext = (deleteSecurityContext_fn)GetProcAddress(_sspi_security_dll, "DeleteSecurityContext");
+
+  // Check if the we managed to map function pointer
+  if(!pfn_deleteSecurityContext) {
+    printf("GetProcAddress failed.\n");
```
niemeyer: That's not a good way to convey an error. GetProcAddr of what, from which library, and definitely not to stdout.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+    );
+
+  // Call the function
+  return status;
+}
+
+/**
+ * Delete Security Context
+ */
+SECURITY_STATUS SEC_ENTRY _sspi_DeleteSecurityContext(PCtxtHandle phContext) {
```
niemeyer: Where is this being used?

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  // Map function
+  #ifdef _UNICODE
+      pfn_acquireCredentialsHandle = (acquireCredentialsHandle_fn)GetProcAddress(_sspi_security_dll, "AcquireCredentialsHandleW");
+  #else
+      pfn_acquireCredentialsHandle = (acquireCredentialsHandle_fn)GetProcAddress(_sspi_security_dll, "AcquireCredentialsHandleA");
+  #endif
+
+  // Check if the we managed to map function pointer
+  if(!pfn_acquireCredentialsHandle) {
+    printf("GetProcAddress failed.\n");
```
niemeyer: That's not a good way to convey an error. GetProcAddr of what, from which library, and definitely not to stdout.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  encryptMessage_fn pfn_encryptMessage = NULL;
+
+  // Return error if library not loaded
+  if(_sspi_security_dll == NULL) return -1;
+
+  // Map function to library method
+  pfn_encryptMessage = (encryptMessage_fn)GetProcAddress(_sspi_security_dll, "EncryptMessage");
+  // Check if the we managed to map function pointer
+  if(!pfn_encryptMessage) {
+    printf("GetProcAddress failed.\n");
```
niemeyer: That's not a good way to convey an error. GetProcAddr of what, from which library, and definitely not to stdout.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+
+static HINSTANCE _sspi_security_dll = NULL; 
+static HINSTANCE _sspi_secur32_dll = NULL;
+
+/**
+ * Encrypt A Message
+ */
+SECURITY_STATUS SEC_ENTRY _sspi_EncryptMessage(PCtxtHandle phContext, unsigned long fQOP, PSecBufferDesc pMessage, unsigned long MessageSeqNo) {
+  // Create function pointer instance
+  encryptMessage_fn pfn_encryptMessage = NULL;
```
niemeyer, line 13: Can we please use the code style (indentation, comment format, etc) as the rest of the code in this package?

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+
+  // Check if the we managed to map function pointer
+  if(!pfn_queryContextAttributes) {
+    printf("GetProcAddress failed.\n");
+    return 42;
+  }
+
+  // Call the function
+  return (*pfn_queryContextAttributes)(
+    phContext, ulAttribute, pBuffer
```
niemeyer: We can have it all in a single line.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+                                          TEXT("InitSecurityInterfaceW"));
+  #else
+    // Get the address of the InitSecurityInterface function.
+    InitSecurityInterface = (INIT_SECURITY_INTERFACE) GetProcAddress (
+                                          _sspi_secur32_dll, 
+                                          TEXT("InitSecurityInterfaceA"));
+  #endif
+
+  if(!InitSecurityInterface) {
+    printf (TEXT("Failed in getting the function address, Error: %x"), GetLastError ());
```
niemeyer: More errors going to stdout. We need a better mechanism for those.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+    return NULL;
+  }
+
+  // Use InitSecurityInterface to get the function table.
+  pSecurityInterface = (*InitSecurityInterface)();
+
+  if(!pSecurityInterface) {
+    printf (TEXT("Failed in getting the function table, Error: %x"), GetLastError ());
```
niemeyer: Ditto.

sasl/kerberos_sspi.c
```diff
@@ -0,0 +1,244 @@
...
+  // Call the function
+  return (*pfn_queryContextAttributes)(
+    phContext, ulAttribute, pBuffer
+  );
+}
+
+/**
+ * InitSecurityInterface
+ */
+PSecurityFunctionTable _ssip_InitSecurityInterface() {
```
niemeyer: Typo in _sspi_.. I don't think we need those underlined prefixes, by the way. The symbol namespace in the package is pretty clean, unless I misunderstand some particularity of Windows.

niemeyer: In fact, is that function even being used?

sasl/sasl_sspi.go
```diff
@@ -0,0 +1,144 @@
+// +build windows
```
niemeyer, line 1: Please call these files sasl_windows.go and sasl_windows.h. No need to have the +build instruction with that.

niemeyer:
Thanks for the work, Valeri.  A few initial comments above.

niemeyer:
Sorry, wrong button.. didn't mean to close it.

sasl/kerberos_sspi.c
vkarpov15, line 2: https://github.com/christkv/kerberos/tree/master/lib/win32 Christian's node driver kerberos code. Its under Apache 2.0.

sasl/sasl_sspi.go
vkarpov15, line 101: Got some weird errors because Close gets called after every step. Figured this would be a reasonable way to not dive down the windows rabbit hole while avoiding leaking memory.

sasl/kerberos_sspi.c
niemeyer, line 2: We need that information at the top of the file itself. Link to the project, who wrote the particular file, what's its license and copyright.

vkarpov15:
In general, I'd rather not make too many changes to `kerberos_sspi.*`, because those are copied directly from Christian's code, and I'd like to be able to import future bug fixes without too much work. I'd appreciate it if you have a better way of pulling in that code though, I agree the style of `kerberos_sspi.*` doesn't match the rest of the code and that's unpleasant.

sasl/sasl_sspi.go
niemeyer, line 101: I don't think I understand the issue. Close cannot be called after a step unless the saslSession is not going to be touched anymore, and if that's the case its resources should be released or we do have leaks. If there are lurking errors, we should understand them as there's likely a fundamental issue in the way we're using the API.

niemeyer:
I'm all for having less code that we need to maintain, but the proper way to do that is to use an external library, which is the process I have followed to implement the current support in Linux, and which hopefully actually works in Windows too (I never got a proper response to whether it does work there or not, and why not, given that the library does mention support for Windows). I'd be quite surprised if there's no proper library that abstract that exact API for multiple operating systems.

That said, if you prefer implementing it that way, and supporting it too, that's okay too. But if we are going down the rabbit role of importing external code, we'll need to maintain that code sane and working for us. That includes file headers, fixing things such as the loose error handling, adding tests, and on the bright side getting rid of code that we're in fact not depending upon. Looks like we can get rid of quite a bit of it.

vkarpov15:
I don't know the exact reasoning why Cyrus SASL doesn't work for Kerberos on Windows. Probably due to the fact that Windows-flavored Kerberos enables you to get a TGT ticket in the process itself without going through `kinit`. As far as I can tell, that's the most obvious difference between these two APIs. Regardless, using SSPI's API on Windows is the established standard for mongodb clients, both the shell and the handful of drivers that implement this use it.

Given the sorry state of go's package management, I think its a better idea to keep the code in there. I'll go in and make kerberos_sspi conform to some reasonable stylistic standards.

niemeyer:
Having experience with other lands, I don't find Go's package management to be in a sorry state, but let's have that conversation in a different forum.

I'm happy to support you in merging this work nevertheless. Let's just try to polish it to a state where we're both happy to maintain it in-tree.

vkarpov15:
Thanks. Let me know if you have any more comments on the most recent iteration, I did what I could to clean up the kerberos_sspi files.

niemeyer:
Going over the comments above, I still see a number of uncovered points.

One easy way to ensure we're on the same page is by responding to the actual comments, even if it's just a _"Done."_  This gives a chance for both of us to tell whether things were completed, or whether there was a disagreement, and what was it.

niemeyer:
Also, can we please drop the "_" prefix from all of the C function names? These are just normal functions in our own namespace.

sasl/sasl_sspi.go
vkarpov15, line 101: Ok done

sasl/kerberos_sspi.c
vkarpov15, line 2: Done

sasl/sasl_sspi.go
vkarpov15, line 1: Done

sasl/kerberos_sspi.c
niemeyer, line 13: This still needs addressing in #25.

sasl/kerberos_sspi.c
niemeyer, line 2: Doesn't look done here or in #25.

sasl/sasl_sspi.go
niemeyer, line 1: Doesn't look done, here or in #25.

PR #25, opened 2014-09-18
POC for Kerberos on Windows with SSPI, take 2

vkarpov15:
Unfortunately can't change branches on a Github PR AFAIK, for the sake of sanity this is #20 against the v2-dev branch

### DIFF
--- a/auth_test.go
+++ b/auth_test.go
@@ -985,7 +985,7 @@ func (kerberosSuite *KerberosSuite) TestAuthKerberosServiceName(c *C) {
 
 	c.Logf("Authenticating with incorrect service name...")
 	err = session.Login(cred)
-	c.Assert(err, ErrorMatches, ".*wrong/ldaptest.10gen.cc@LDAPTEST.10GEN.CC not found.*")
+	c.Assert(err, ErrorMatches, ".*@LDAPTEST.10GEN.CC not found.*")
 
 	cred.Service = rightServiceName
 	c.Logf("Authenticating with correct service name...")


niemeyer:
There are comments in #20 that were not addressed. Even some of the ones marked as "Done".

vkarpov15:
Yeah I hadn't pushed them yet, sorry. Subsequent commit contains them. I just wanted to open up a new PR because I merged v2-dev into this branch and so the previous PR had a lot of noise.

niemeyer:
Okay, please let me know once this is actually ready for review.

vkarpov15:
Go ahead and review, I'm reasonably certain I've addressed all your comments.

sasl/sasl.go
```diff
@@ -2,6 +2,9 @@
 //
 // This package is not meant to be used by itself.
 //
+
+// +build linux darwin
```
niemeyer: // +build !windows

sasl/sasl_windows.go
```diff
@@ -0,0 +1,142 @@
...
+	return cstr
+}
+
+func (ss *saslSession) Close() {
+	for _, cstr := range ss.stringsToFree {
+		C.free(unsafe.Pointer(cstr))
+	}
+
+	// Make sure we've cleaned up all the buffers we malloced when we're sure we don't need em anymore
+	if ss.authComplete || ss.errored {
```
niemeyer: This was discussed in the prior review, and still looks pending.

sasl/sasl_windows.c
```diff
@@ -0,0 +1,126 @@
+#ifdef _WIN32
```
niemeyer: I don't think this file will get built when not on windows, given its name.

sasl/sasl_windows.h
```diff
@@ -0,0 +1,9 @@
+#ifdef _WIN32
```
niemeyer: I don't think this file will get built when not on windows, given its name.

niemeyer:
Ok, just did a quick scan with a few comments, but I'll need to go back to it later.

Meanwhile, where are the tests?

vkarpov15:
Thanks for comments, I'll fix them in a bit. Re: tests, working on that, that will require moving kerb tests to a separate suite for now. My current testing setup overwrites existing tests and so isn't mergable.

sasl/sasl_windows.h
vkarpov15: done

sasl/sasl_windows.c
vkarpov15: done

sasl/sasl_windows.go
vkarpov15: done

sasl/sasl.go
vkarpov15: done

vkarpov15:
Hey Gustavo, just a gentle reminder to see if you have any more comments.

niemeyer:
I might, but where are the tests?

vkarpov15:
In a separate branch and ready to merge into this one. Do you want them in
this branch? There's some significant work there to break Kerberos tests
off into their own suite so they don't go through the normal setup and
teardown.

On Mon, Sep 22, 2014 at 12:30 PM, Gustavo Niemeyer <notifications@github.com

niemeyer:
I don't understand what kind of significant work might take place for that. Registering a new test suite requires three or four lines of code.

Either way, it doesn't make sense to merge this without tests. Please also make sure that the tests that are being submitted with the logic actually exercise it well, and are passing for real.

niemeyer:
Also, what do you mean by "break Kerberos tests"?  The test that exists today are trivial in their implementation and design. We shouldn't be breaking them out or removing them from where they are.

niemeyer:
It might be best to submit that new logic you're talking about in a separate branch before polluting this PR, actually, so I can at least have an idea about what's being done.

vkarpov15:
See [here](https://github.com/vkarpov15/mgo/pull/1) for diff. The general idea is:

1) have a separate test suite, `RemoteSuite`, for tests that don't require a local mongod setup. With this, the Kerberos tests no longer have to use the `Setup()` and `TearDown`, so the tests can run on Windows.
2) I switched to using the official drivers Kerberos testing host instead of the mms one, because its the standard practice, and SSPI doesn't really do keytabs as far as I know.
3) Because SSPI requires you to acquire your kerberos ticket in-process, I added an environment variable for setting the password that gets passed to SSPI

niemeyer:
Okay, that looks reasonable, but let's name it what it is: KerberosSuite, and have the suite defined right above kerberosFlag instead of having a new file for it. We can reorganize it after it lands if necessary.

If that's what it is, we can merge it in this PR so you can make sure tests pass before we merge this.

vkarpov15:
Ok I merged in the test suite. Here's a screen cap of the windows tests passing:

image

niemeyer:
Thanks! Can you please run the tests on Linux as well, to make sure we're still good there after the changes?

vkarpov15:
Tests now pass on ubuntu 12.04 as well:

image

vkarpov15:
Hey Gustavo, just a ping for any more comments

PR #27, opened 2014-09-23
MONGODB-X509 authentication

gabrielrussell:
Gustavo,

This is what I need to do to make MONGODB-X509 authentication to work. I'm happy to make whatever changes you deem necessary. Have a great day.

Gabriel


niemeyer:
Thanks for this, Gabriel. Looks quite nice. I'll add a comment inline, but the main thing this is missing is a test. This logic can be exercised by default, right? If so, we need a clean test that always run with it.

auth.go
```diff
@@ -44,6 +44,12 @@ type authCmd struct {
 	Key   string
 }
 
+type authX509Cmd struct {
```
niemeyer: Would you mind to move this to right above the function where it is used?

gabrielrussell:
Hey Gustavo. Here is a test, along with a couple more changes to the driver to get the tests to work. Let me know what you think.

PR #29, opened 2014-09-29
Fix SSPI include path

vkarpov15:
Hey Gustavo,

Your changes for #25 look good, you just forgot to change one include path. With this change, KerberosSuite passes on Windows.


niemeyer:
Thanks!

PR #30, opened 2014-09-30
Add build tag to make sure sasl.c isn't built on windows

vkarpov15:


vkarpov15:
Also fixed an [incorrect authentication mechanism name](http://docs.mongodb.org/manual/core/authentication/#authentication-mechanisms)

niemeyer:
Thanks for the build fix, happy to apply it.

That said, we cannot apply the suggested rename as-is. People out there are likely depending on this name, and it was modified in the server right before the 2.4 release:

```
https://jira.mongodb.org/browse/SERVER-8501
```

We should support the two names instead.

To avoid another roundtrip, I have applied the two fixes to v2-dev. Thanks again for the changes.

vkarpov15:
@niemeyer I unfortunately don't see the MONGODB-CR change in v2-dev. Would go a long way to getting our tools back on the main driver and off of our fork :)

niemeyer:
Sorry, it was late and I forgot to push it. Should be there now.

PR #31, opened 2014-10-01
Support for DBPointer in bson package

mpobrien:
According to bsonspec.org, this is a deprecated type, however there are some mongodump/mongorestore tests that rely on encoding and decoding of objects which contain it. This commit adds support for the DBPointer type and includes a test for it.

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -386,6 +386,14 @@ type JavaScript struct {
 	Scope interface{}
 }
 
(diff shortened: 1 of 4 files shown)


niemeyer:
Thanks Mike.

niemeyer:
I have merged this to v2 by mistake, and quickly reverted back.

This is now properly merged in v2-dev with 4d7c7241 and will be in the next release.

PR #32, opened 2014-10-06
Support struct IDs

waigani:
Support txns of docs with struct IDS.

### DIFF
--- a/txn/txn.go
+++ b/txn/txn.go
@@ -7,13 +7,15 @@
 package txn
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
-	"gopkg.in/mgo.v2"
-	"gopkg.in/mgo.v2/bson"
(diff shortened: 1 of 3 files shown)


txn/txn.go
```diff
@@ -483,10 +485,25 @@ func (ks docKeys) Less(i, j int) bool {
...
 	case natureBool:
 		return !av.(bool) && bv.(bool)
+	case natureStruct:
+		return structLess(av, bv)
 	}
 	panic("unreachable")
 }
 
+func structLess(av, bv interface{}) bool {
+	avm, err := bson.Marshal(av)
```
niemeyer, line 495: In the worst case of a quicksort that'll be N^2 marshalings, for each struct!

mjs, line 495: You're right - that's not good. See #34 for an alternative approach.

PR #33, opened 2014-10-06
Quick fix for Kerberos: don't use hardcoded userPlusRealm length

vkarpov15:
Hey Gustavo,

Looks like I made a silly mistake in my SSPI code, forgot to take out the hardcoded username length.

### DIFF
--- a/sasl/sasl_windows.c
+++ b/sasl/sasl_windows.c
@@ -102,6 +102,7 @@ int sspi_send_client_authz_id(CtxtHandle *context, PVOID *buffer, ULONG *buffer_
 
 	status = call_sspi_encrypt_message(context, SECQOP_WRAP_NO_ENCRYPT, &wrapBufDesc, 0);
 	if (status != SEC_E_OK) {
+		free(msg);
 		return status;
 	}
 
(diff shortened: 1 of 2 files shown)


sasl/sasl_windows.c
```diff
@@ -73,13 +73,14 @@ int sspi_send_client_authz_id(CtxtHandle *context, PVOID *buffer, ULONG *buffer_
 		return status;
 	}
 
-	int msgSize = 4 + 25;
+	size_t user_plus_realm_length = strlen(user_plus_realm);
+	int msgSize = 4 + user_plus_realm_length;
 	char *msg = malloc((sizes.cbSecurityTrailer + msgSize + sizes.cbBlockSize) * sizeof(char));
```
niemeyer, line 78: There seems to be a leak of this allocation. I suggest dropping the memcpy below, and instead simply doing:

```
*buffer = msg;
*buffer_length = ...
```

so that you pass the reference back into the Go code rather than losing it.

Then, the Go code that calls this also seems a bit problematic. It should probably do something alone the lines of:

```
if buffer != C.PVOID(nil) {
        defer C.free(unsafe.Pointer(buffer))
}
```

This will avoid the need for buffersToFree, and most importantly will not crash the whole process when there are any errors in send_client_authz_id that prevent it from sending a proper buffer. Note the defer.. this allows handling the deallocation early on, and then forgetting about it in a good sense.

Note that GoBytes will copy the data, so it is okay to drop the underlying data.

vkarpov15, line 78: You're right about msg getting leaked, but I think it's simpler to just `free()` it when you're done with it. That way the current structure works as intended.

niemeyer, line 78: Sure, the process of re-allocating and re-copying the data looks suspect, but I'll leave that up to you.

That said, let's please fix the Go side of this. That buffersToFree list is completely unnecessary, as suggested in the comment above, and it's also going to crash the application if there are errors, since the buffer is not allocated. It should be trivial to free with defer as suggested.

vkarpov15, line 78: done

niemeyer:
Thanks for the changes.

PR #34, opened 2014-10-07
Improved document key sorting and support for struct document ids

mjs:
Removed unnecessary resort of document keys

---

Introduce sortableDocKey

Precompute the nature and sort value for each doc key once instead of each time Less is called. This trades slight additional complexity and memory usage for sort speed. This also paves the way for sorting of struct document keys.

---

Support sorting of struct based document ids

### DIFF
--- a/txn/debug.go
+++ b/txn/debug.go
@@ -70,12 +70,12 @@ func argsForLog(args []interface{}) []interface{} {
 			args[i] = lst
 		case map[docKey][]bson.ObjectId:
 			buf := &bytes.Buffer{}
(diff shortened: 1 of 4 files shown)


niemeyer:
That's improving the previous approach with caching, but the underlying concept is still not good. There's no reason to pay the cost of marshaling of these structs simply for obtaining a well defined order. It's also pretending to work for things that actually do not work. Marshal will marshal anything, including things with an undefined order. These should continue to break.

We should be able to easily have an ordering mechanism for the structs we care about without resorting to marshaling.

PR #35, opened 2014-10-07
Sort struct dockeys

waigani:


txn/flusher.go
```diff
@@ -225,10 +224,10 @@ func (f *flusher) prepare(t *transaction, force bool) (revnos []int64, err error
 	}
 	f.debugf("Preparing %s", t)
 
-	// Iterate in a stable way across all runners. This isn't
-	// strictly required, but reduces the chances of cycles.
+	// docKeys is sorted to support stable iteration across all
```
niemeyer: s/docKeys is sorted to support/dkeys being sorted means/

txn/flusher.go
```diff
@@ -380,10 +379,10 @@ func (f *flusher) rescan(t *transaction, force bool) (revnos []int64, err error)
 		panic(fmt.Errorf("rescanning transaction in invalid state: %q", t.State))
 	}
 
-	// Iterate in a stable way across all runners. This isn't
-	// strictly required, but reduces the chances of cycles.
+	// docKeys is sorted to support stable iteration across all
```
niemeyer: s/docKeys is sorted to support/dkeys being sorted means/

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+	if a.C != b.C {
+		return a.C < b.C
+	}
+	return valuecmp(a.Id, b.Id) == -1
+}
+
+func valuecmp(a, b interface{}) int {
+	av, an := valueNature(a)
+	bv, bn := valueNature(b)
+	if n := intcmp(int64(an), int64(bn)); n != 0 {
```
niemeyer: Many of the changes seem to abstract away what used to be straightforward statements in the original algorithm, without great benefit. The code above, for example, was:

```
if an != bn { return an < bn }
```

and we can preserve it pretty much as simple:

```
if bn > an {
        return 1
}
if an < bn {
        return -1
}
```

then, we don't need all the sub-functions below either. Instead, we can do:

```
if av == bn {
        return 0
}
var less bool
switch an {
case natureString:
        less = av.(string) < bv.(string)
...
}
if less {
        return -1
}
return 1
```

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+		return boolcmp(av.(bool), bv.(bool))
+	case natureStruct:
+		return structcmp(av, bv)
+	}
+	panic("unreachable")
+}
+
+func structcmp(a, b interface{}) int {
+	at := reflect.TypeOf(a)
+	bt := reflect.TypeOf(b)
```
niemeyer: Please move these below av/bv, and do:

```
at := av.Type()
bt := bv.Type()
```

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+	av := reflect.ValueOf(a)
+	bv := reflect.ValueOf(b)
+	if n := intcmp(int64(at.NumField()), int64(bt.NumField())); n != 0 {
```
niemeyer: That's not right. Please drop it.

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+		return n
+	}
+	for i := 0; i < at.NumField(); i++ {
```
niemeyer: Needs to take into account whether the field is exported or not.

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+		nameA := getFieldName(at.Field(i))
+		nameB := getFieldName(bt.Field(i))
+		if n := strcmp(nameA, nameB); n != 0 {
+			return n
+		}
+		if n := valuecmp(av.Field(i).Interface(), bv.Field(i).Interface()); n != 0 {
```
niemeyer: Please compare the value first. In the vast majority of cases this will be the delta, and it'll never reach the more complex field-comparing logic.

txn/dockey.go
```diff
@@ -0,0 +1,114 @@
...
+			return n
+		}
+	}
+	return 0
+}
+
+func getFieldName(f reflect.StructField) string {
+	name := f.Tag.Get("bson")
+	s := strings.Split(name, ",")
```
niemeyer: Please find the prefix instead of splitting:

```
if i := strings.Index(name, ","); i >= 0 {
        name = name[:i]
}
```

that's a trivial, non-allocating, and to-the-point operation compared to splitting.

txn/dockey_test.go
```diff
@@ -0,0 +1,244 @@
...
+	TestingT(t)
+}
+
+type DocKeySuite struct{}
+
+var _ = Suite(&DocKeySuite{})
+
+func (s *DocKeySuite) TestInt(c *C) {
+	check(c,
+		docKeys{
```
niemeyer: Thanks for the tests, but please move all of these to a table structure on a single test slice, and run it from a single test function:

```
var docKeysTests = [][]docKeys{{
    {
        {"c", 1},
        {"c", 5},
        {"c", 2}
    }, {
        {"c", 1},
        {"c", 2},
        {"c", 5}
    },
}, {
     ...
}}
```

waigani:
PTAL

txn/dockey.go
```diff
@@ -0,0 +1,117 @@
...
+		minField = bt.NumField()
+	}
+
+	for i := 0; i < minField; i++ {
+		if n := valuecmp(av.Field(i).Interface(), bv.Field(i).Interface()); n != 0 {
+			return n
+		}
+		nameA := getFieldName(at.Field(i))
+		nameB := getFieldName(bt.Field(i))
+		if n := bytes.Compare([]byte(nameA), []byte(nameB)); n != 0 {
```
niemeyer:
```
if nameA < nameB {
        return -1
}
if nameA > nameB {
        return 1
}
```

txn/dockey.go
```diff
@@ -0,0 +1,117 @@
...
+			return n
+		}
+	}
+	if af < bf {
+		return -1
+	} else if af > bf {
+		return 1
+	}
```
niemeyer: Consider these two structs:

```
type A struct {
        S string
}
type B struct {
        p, q, r bool
        S string
}
```

txn/dockey.go
```diff
@@ -0,0 +1,117 @@
...
+	return 0
+}
+
+func numExportedField(t reflect.Type) int {
```
niemeyer: Please drop this logic. It's not helping. See above.

txn/dockey_test.go
```diff
@@ -0,0 +1,176 @@
...
+	type T10 struct {
+		A int    `bson:"c"`
+		B string `bson:"d,omitempty"`
+	}
+	type T11 struct {
+		C int
+		D string
+	}
+
+	for i, test := range [][]docKeys{{
```
niemeyer: Please move these out to a top level variable named `docKeysTests`.

waigani:
PTAL

txn/dockey.go
```diff
@@ -0,0 +1,118 @@
...
+
+	var minField int
+	if afl < bfl {
+		minField = afl
+	} else {
+		minField = bfl
+	}
+
+	for i := 0; i < minField; i++ {
+		if n := valuecmp(av.FieldByName(af[i].Name).Interface(), bv.FieldByName(bf[i].Name).Interface()); n != 0 {
```
niemeyer: This builds two lists with all exported fields, potentially not entirely used, and then dispatches a search by name to the reflect machinery on every single field. I think we can do better with an approach along these lines:

http://paste.ubuntu.com/8526037/

I coded that blindly (not even compiled). Can you please try to make it work?

If it does work, we're ready, but before merging, please move the content back where it was. It's just adding a couple of small functions now, and this file does not justify itself anymore.

waigani:
PTAL, keeping dockey_test.go as an internal test because docKeys is internal to txn.

txn/txn.go
```diff
@@ -513,6 +484,121 @@ func valueNature(v interface{}) (value interface{}, nature typeNature) {
...
+	av := reflect.ValueOf(a)
+	bv := reflect.ValueOf(b)
+
+	var ai, bi = 0, 0
+	var an, bn = av.NumField(), bv.NumField()
+	var af, bf reflect.StructField
+	for {
+		for ai < an {
+			af = av.Type().Field(ai)
+			ai++
```
niemeyer: It looks like the logic is now incorrect. Consider that av.Field(ai) below will see something that is not af, and if ai == an below, the loop will leave without ever having looked at af, which means both the comparison is happening across unexpected fields, and that we're missing fields entirely. It's slightly worrying that we have no tests breaking.

waigani:
PTAL, Added failing test which now passes.

PR #36, opened 2014-10-08
MONGODB-X509 + unit tests

gabrielrussell:
Gustavo,

I had to make two changes to get my test to run. One is that OtherDBRoles should be usable with "$external" as well as "admin". The other is that, in order to enable the "localhost exception" if updateUser fails with an auth error (13) we need to try the createUser. I'm fairly certain that the "localhost exception" fix is correct, but I'd like validation from you that that the $external db OtherDBRoles change is correct.

I have an additional change where I replace the low valued mongod response codes with constants, but I'll submit that separately, because I think you might just not like it.

Thanks,
Gabriel


niemeyer:
Thanks for that. I don't have energy right now to review this, but I'll try to get over it in the next couple of days, or over the weekend in the worst case.

niemeyer:
The change looks good overall, thank you.

There are trivial tweaks to be made, such as ensuring that the tests run whenever the server at hand is SSL-capable, instead of requiring manual procedures that imply the test would be rarely run.

I'll merge this right now, though, and will do these tweaks myself afterwards. Thanks again.

niemeyer:
Okay, I performed the mentioned tweaks, and also fixed the test so that it works if run multiple times (second time would fail to add the Admin user because it had no auth). That was done by integrating the new server in the common init and cleanup workflow, whenever SSL is available.

You can see all those tweaks in c1b9c9f28. Then, the auth test was moved to the common auth test file in dc6fe4aaa.

Please let me know if you have any questions.

PR #37, opened 2014-10-08
constant defs for low-valued command result codes

gabrielrussell:
Gustavo,

I only did the low values because the high values don't all have names and are not even really guaranteed to remain constant.

Some of the high values have names and are probably not going to change, and I could add them to this if you like.

Thanks
Gabriel


niemeyer:
Hey Gabriel,

I wonder if we should have these in. In my experience the error codes are quite unreliable, changed over time, and there are often multiple codes for the same error (see IsDup for an idea). Being just partial also hints of suboptimal value.

It feels like it might be worth continuing with the pattern we have at the moment: important errors are either represented by first time constants (ErrNotFound) or have comprehensive checkers (IsDup). Other errors are easy for people to have their own verifiers or code constants in the local code.

PR #40, opened 2014-10-21
Bulk operations: add Count method

julien-c:


niemeyer:
Would you mind to describe in more detail what the problem being solved is?

There are still questions about how the internal implementation of bulk handling will look like, and what the meaning of the count method would be. There is the potential for grouping operations, for example, so would that mean "count the number of requests" or "count the number of individual changes".

For those reasons, it would probably be best to wait until the Bulk API is fully fleshed out before we add such introspection methods.

klauspost:
Count() would be nice, because sending a bulk request with 0 documents gives an error from Mongo.

So for flushing purposes it would be nice to be able to check for 0 documents, or otherwise Run() should just return if the count is 0.

julien-c:
I think this is the reason I needed it myself.

niemeyer:
Thanks, I'll consider this further once the bulk API is completed. For now, unfortunately I'll need to delete this pull request because I need to rename the v2-dev branch to enable us to make it visible via gopkg.in (it needs to be -unstable), and unfortunately GitHub doesn't offer any way to rename a branch politely.

julien-c:
No problem. I'll keep an eye open for the Bulk API.

PR #41, opened 2014-10-29
add support for repairCursor command with test coverage

mpobrien:
A new command called "repairCursor" was added to the server in 2.7.8 (see https://jira.mongodb.org/browse/SERVER-15544 for details). We need support for this in the mgo driver so that tools which involve database repair operations (like mongodump) can access it. 
It's implemented as a command which returns a cursor; the server performs a best-effort attempt to recover any data from damaged data files and return the documents it finds on the cursor - so, it could potentially return multiple copies of the same document. I've written the test to account for this behavior.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1881,6 +1881,11 @@ func (c *Collection) Pipe(pipeline interface{}) *Pipe {
 	}
 }
 
+// Repair calls the 'repairCursor' command (supported in mongo 2.7.8 and later)
+// and returns an iterator to go through the results.
+// This command requests the server to perform a best-effort attempt to recover
+// all documents from the collection in cases of damaged data files, so it may
+// return multiple copies of the same document.
 func (c *Collection) Repair() *Iter {
 	// Clone session and set it to strong mode so that the server
 	// used for the query may be safely obtained afterwards, if
--- a/session.go
+++ b/session.go
@@ -1845,6 +1845,15 @@ type pipeCmdCursor struct {
 	BatchSize int `bson:"batchSize,omitempty"`
 }
 
+type repairCmd struct {
+	RepairCursor string           `bson:"repairCursor"`
+	Cursor       *repairCmdCursor ",omitempty"
+}
+
+type repairCmdCursor struct {
+	BatchSize int `bson:"batchSize,omitempty"`
+}
+
 // Pipe prepares a pipeline to aggregate. The pipeline document
 // must be a slice built in terms of the aggregation framework language.
 //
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -1872,6 +1881,60 @@ func (c *Collection) Pipe(pipeline interface{}) *Pipe {
 	}
 }
 
+func (c *Collection) Repair() *Iter {
```
niemeyer, line 1890: Please provide user-oriented documentation. The other methods are a good example for the style in which it should be worded.

Also, is there any documentation at the docs site for what the command does?

session.go
```diff
@@ -1872,6 +1881,60 @@ func (c *Collection) Pipe(pipeline interface{}) *Pipe {
...
+	// Clone session and set it to strong mode so that the server
```
niemeyer, line 1891: The amount of boilerplate in this method, which is a copy of the aggregation equivalent, is unfortunate. Let's move forward with it, though, and I'll refactor into something more appropriate later.

session_test.go
```diff
@@ -3192,6 +3192,48 @@ func (s *S) TestFsync(c *C) {
...
+	resultCounts := map[int]int{}
+	for repairIter.Next(&result) {
+		resultCounts[result.N]++
+	}
+
+	c.Assert(repairIter.Next(&result), Equals, false)
+	c.Assert(repairIter.Err(), IsNil)
+	c.Assert(repairIter.Close(), IsNil)
+
+	/* Verify that the results of the repair cursor are valid.
```
niemeyer: Please use the // style of comments, per the rest of the code.

session.go
mpobrien, line 1890: OK - added some docstring info to this method. 
Unfortunately there's no information about this at the docs site since it's new feature, but I can look into getting it added there and then update this section when there's something I can link to.

session_test.go
mpobrien: Done.

session.go
mpobrien, line 1891: I agree. Refactoring out the boilerplate in a way that also affects the functions for aggregation seemed a bit intrusive for me to attempt within this pull request, so I figured I would defer that piece to you.
Perhaps a more generic way to run commands that return cursors (which both Pipe() and this command could leverage) would be a good thing to add in the future.

session.go
```diff
@@ -1872,6 +1881,65 @@ func (c *Collection) Pipe(pipeline interface{}) *Pipe {
 	}
 }
 
+// Repair calls the 'repairCursor' command (supported in mongo 2.7.8 and later)
+// and returns an iterator to go through the results.
```
niemeyer: Thanks, but I'd prefer to avoid mentioning implementation details. We don't do that in any other case.

Here is a suggested version, based on yours:

// Repair returns an iterator that goes over all recovered documents in the collection,
// in a best-effort manner. This is most useful when there are damaged data files.
// Multiple copies of the same document may be returned by the iterator.
//
// Repair is supported in MongoDB 2.7.8 and later.

(update: included the MongoDB version)

mpobrien: Sounds good. Updated to use your suggested text.

mpobrien:
Hey Gustavo, sorry but just wanted to check in on this - any other changes you'd like to see here?

niemeyer:
Hey Mike,

Sorry for the feedback. This is looking good, and was already in my agenda for merging today. I'm going to merge it directly to v2 since it feels independent enough to not cause any harm to people that are not depending on the feature yet. Hopefully this will give you a hand there.

mpobrien:
Thanks Gustavo!

PR #51, opened 2014-11-24
Enable replica set name assertion in server discovery

deafgoat:
Hi Gustavo,

The changes here are meant to address [issue 45](https://github.com/go-mgo/mgo/issues/45). In summary:
- The "DialInfo" struct now accepts a "ReplicaSetName" field that if non-empty, is used to ensure all nodes found during discovery belong to the provided replica set (this is used for sessions established with the "DialWithInfo" function).
- The "DialWithTimeout" function is also updated to check for the replica set name and set the"ReplicaSetName" field.
- Refactored a bit of the tests to check for "ErrNoReachableServers" instead of a string literal.

Important details
- For session establishment, an error is returned if no seed node is part of the replica set. If at least one node is part of the replica set, other non-member seeds are discarded. The usual mechanisms around discovery work as before.
- If no seed node is part of the replica set, an error is returned on dial.
- If no "ReplicaSetName" is provided, session establishment works as before - with the assumption that the servers supplied are either individual mongod servers or mongos routers.

Please take a look and let me know if anything is amiss.

Thanks,
Wisdom.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -54,16 +54,11 @@ type mongoCluster struct {
 	direct       bool
 	failFast     bool
 	syncCount    uint
-	err          error
 	setName      string
 	cachedIndex  map[string]bool
 	sync         chan bool
 	dial         dialer
 }
-
-var (
-	ErrNoReachableServers = errors.New("no reachable servers")
-)
 
 func newCluster(userSeeds []string, direct, failFast bool, dial dialer, setName string) *mongoCluster {
 	cluster := &mongoCluster{
@@ -208,8 +203,7 @@ func (cluster *mongoCluster) syncServer(server *mongoServer) (info *mongoServerI
 
 	if cluster.setName != "" && result.SetName != cluster.setName {
 		log("SYNC Server ", addr, " not a member of replica set ", cluster.setName)
-		cluster.err = errors.New(addr + " is not part of " + cluster.setName + " replica set")
-		return nil, nil, cluster.err
+		return nil, nil, errors.New(addr + " is not part of " + cluster.setName + " replica set")
 	}
 
 	if result.IsMaster {
@@ -270,9 +264,8 @@ func (cluster *mongoCluster) addServer(server *mongoServer, info *mongoServerInf
 			cluster.Unlock()
 			server.Close()
 			return
-		} else {
-			cluster.servers.Add(server)
 		}
+		cluster.servers.Add(server)
 		if info.Master {
 			cluster.masters.Add(server)
 			log("SYNC Adding ", server.Addr, " to cluster as a master.")
@@ -561,10 +554,7 @@ func (cluster *mongoCluster) AcquireSocket(slaveOk bool, syncTimeout time.Durati
 				syncCount = cluster.syncCount
 			} else if syncTimeout != 0 && started.Before(time.Now().Add(-syncTimeout)) || cluster.failFast && cluster.syncCount != syncCount {
 				cluster.RUnlock()
-				if cluster.err != nil {
-					return nil, cluster.err
-				}
-				return nil, ErrNoReachableServers
+				return nil, errors.New("no reachable servers")
 			}
 			log("Waiting for servers to synchronize...")
 			cluster.syncServers()
--- a/cluster.go
+++ b/cluster.go
@@ -202,8 +202,8 @@ func (cluster *mongoCluster) syncServer(server *mongoServer) (info *mongoServerI
 	}
 
 	if cluster.setName != "" && result.SetName != cluster.setName {
-		log("SYNC Server ", addr, " not a member of replica set ", cluster.setName)
-		return nil, nil, errors.New(addr + " is not part of " + cluster.setName + " replica set")
+		log("SYNC Server ", addr, " is not a member of replica set ", cluster.setName)
+		return nil, nil, errors.New(addr + " is not a member of replica set " + cluster.setName)
 	}
 
 	if result.IsMaster {
@@ -257,12 +257,6 @@ func (cluster *mongoCluster) addServer(server *mongoServer, info *mongoServerInf
 			cluster.Unlock()
 			server.Close()
 			log("SYNC Discarding unknown server ", server.Addr, " due to partial sync.")
-			return
-		}
-		if cluster.setName != "" && info.SetName != cluster.setName {
-			log("SYNC Discarding ", server.Addr, " not part of ", cluster.setName, " replica set.")
-			cluster.Unlock()
-			server.Close()
 			return
 		}
 		cluster.servers.Add(server)
--- a/cluster.go
+++ b/cluster.go
@@ -54,18 +54,25 @@ type mongoCluster struct {
 	direct       bool
 	failFast     bool
 	syncCount    uint
+	err          error
+	setName      string
 	cachedIndex  map[string]bool
 	sync         chan bool
 	dial         dialer
 }
 
-func newCluster(userSeeds []string, direct, failFast bool, dial dialer) *mongoCluster {
+var (
+	ErrNoReachableServers = errors.New("no reachable servers")
+)
+
+func newCluster(userSeeds []string, direct, failFast bool, dial dialer, setName string) *mongoCluster {
 	cluster := &mongoCluster{
 		userSeeds:  userSeeds,
 		references: 1,
 		direct:     direct,
 		failFast:   failFast,
 		dial:       dial,
+		setName:    setName,
 	}
 	cluster.serverSynced.L = cluster.RWMutex.RLocker()
 	cluster.sync = make(chan bool, 1)
@@ -131,7 +138,8 @@ type isMasterResult struct {
 	Passives       []string
 	Tags           bson.D
 	Msg            string
-	MaxWireVersion int `bson:"maxWireVersion"`
+	SetName        string `bson:"setName"`
+	MaxWireVersion int    `bson:"maxWireVersion"`
 }
 
 func (cluster *mongoCluster) isMaster(socket *mongoSocket, result *isMasterResult) error {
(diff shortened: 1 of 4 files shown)


mpobrien:
Hey Gustavo - any further thoughts on this one or https://github.com/go-mgo/mgo/issues/50? It would be really useful for us to fix some rough edges in the new mongo tools.

niemeyer:
The change looks great. There are only a few minors which I will comment inline.

cluster.go
```diff
@@ -54,18 +54,25 @@ type mongoCluster struct {
...
+	err          error
+	setName      string
 	cachedIndex  map[string]bool
 	sync         chan bool
 	dial         dialer
 }
 
-func newCluster(userSeeds []string, direct, failFast bool, dial dialer) *mongoCluster {
+var (
+	ErrNoReachableServers = errors.New("no reachable servers")
```
niemeyer: Please do not change that in this branch. See #50 for discussion on it.

cluster.go
```diff
@@ -250,7 +265,14 @@ func (cluster *mongoCluster) addServer(server *mongoServer, info *mongoServerInf
 			log("SYNC Discarding unknown server ", server.Addr, " due to partial sync.")
 			return
 		}
-		cluster.servers.Add(server)
+		if cluster.setName != "" && info.SetName != cluster.setName {
+			log("SYNC Discarding ", server.Addr, " not part of ", cluster.setName, " replica set.")
+			cluster.Unlock()
+			server.Close()
+			return
+		} else {
```
niemeyer: Please drop the else line. The above statement scope block is returning, so it's unnecessary logic making the code less clear.

cluster.go
```diff
@@ -539,7 +561,10 @@ func (cluster *mongoCluster) AcquireSocket(slaveOk bool, syncTimeout time.Durati
 				syncCount = cluster.syncCount
 			} else if syncTimeout != 0 && started.Before(time.Now().Add(-syncTimeout)) || cluster.failFast && cluster.syncCount != syncCount {
 				cluster.RUnlock()
-				return nil, errors.New("no reachable servers")
+				if cluster.err != nil {
```
niemeyer: Race condition here. Must be read within the lock.

niemeyer:
That's about it. A few entries, but nothing fundamental. Thanks for these changes.

niemeyer:
While thinking about that change further, there is a more fundamental issue. The cluster.err variable is propagated on any error, irrespective of whether the failure happened with that specific server. Now imagine there are 5 servers in a replica set. If one of the servers turns out to be misconfigured, it's nice to reject communication with it, but it would be bogus to fail against any other correct server with that error.

Makes sense?

deafgoat:
Hi Gustavo,

Thanks for taking a look. I'm holding off on updating the code to address the first set of issues you raised since I don't completely understand your last comment. The cluster.err variable is only set when a server in the seed list - or during the course of discovery - is found not to be part of the requested replica set. Once set, the error is only propagated insofar as [spawnSync](https://github.com/go-mgo/mgo/blob/4890377a975625a156d530718c118267c5256f4b/cluster.go#L509) for that server is short-circuited (so communication with that server is rejected). The other goroutines for possibly properly configured servers proceed as usual.

By [not adding](https://github.com/go-mgo/mgo/blob/4890377a975625a156d530718c118267c5256f4b/cluster.go#L484) it to the list of known servers, there can be no communication with it. While cluster.err is still set at that point, it is only used if the code gets [here](https://github.com/go-mgo/mgo/blob/4890377a975625a156d530718c118267c5256f4b/cluster.go#L563). At this point, we only use cluster.err - if it's set - to indicate a possible cause of the issue (note that at this point, we're already returning "no reachable servers"). Do you mean we should not assume cluster.err at that point? Would it suffice to just remove the check for cluster.err [here](https://github.com/go-mgo/mgo/blob/4890377a975625a156d530718c118267c5256f4b/cluster.go#L564-L566)?

niemeyer:
> Once set, the error is only propagated insofar as spawnSync for that server is short-circuited (so communication with that server is rejected). The other goroutines for possibly properly configured servers proceed as usual.

That's not quite true. Consider which code paths can lead to the return of cluster.err, and what is the relationship of the call site with the bad server.

deafgoat:
OK, I'll completely remove cluster.err.

deafgoat:
Hey Gustavo, removed the use of cluster.err. Please let me know if there are any other issues I need to address, thanks.

niemeyer:
Thanks!

niemeyer:
The change is looking good, and I'm happy to merge it. There's one detail which would be good to address before we do that: some of these tests are now quite ineffective at testing anything at all, by just dialing and making sure there are no errors. There are good ways to verify that the offending server is not being communicated with:
1. You can get the set of current live servers from session.LiveServers()
2. You can inspect the log to verify that the expect error was reported, via c.GetTestLog() in the test method

Would you mind to improve the tests a little so we can be sure they're exercising what they seem to be?

deafgoat:
Done.

cluster.go
```diff
@@ -250,6 +259,12 @@ func (cluster *mongoCluster) addServer(server *mongoServer, info *mongoServerInf
 			log("SYNC Discarding unknown server ", server.Addr, " due to partial sync.")
 			return
 		}
+		if cluster.setName != "" && info.SetName != cluster.setName {
```
niemeyer: Why do we need this test?  It looks like the only way we can possibly add a server is by syncing it, which will error out if the server is not in the valid replica set. What am I missing here?

deafgoat: You're right, we'll never get here with a server we shouldn't accept.

deafgoat:
Removed the superfluous check.

cluster_test.go
```diff
@@ -907,6 +907,230 @@ func (s *S) TestSocketTimeoutOnInactiveSocket(c *C) {
...
+
+	runTest := func(session *mgo.Session, err error) {
+		c.Assert(err, IsNil)
+		defer session.Close()
+
+		coll := session.DB("mydb").C("mycoll")
+		coll.Insert(M{"a": 1, "b": 2})
+
+		result := struct{ Ok bool }{}
+		err = session.Run("getLastError", &result)
```
niemeyer, line 927: I'm merging it, and will do the remaining changes, but I'll add a few extra notes here just for awareness.

In the line above, rather than re-issuing a getLastError, the result of Insert should be checked.

cluster_test.go
```diff
@@ -907,6 +907,230 @@ func (s *S) TestSocketTimeoutOnInactiveSocket(c *C) {
...
+	connectionUrl := "mongodb://localhost:40022/?replicaSet=rs1"
+	_, err = mgo.Dial(connectionUrl)
+	c.Assert(err, ErrorMatches, "no reachable servers")
+
+	connectionUrl += "&connect=direct"
+	_, err = mgo.Dial(connectionUrl)
+	c.Assert(err, ErrorMatches, "no reachable servers")
+}
+
+func (s *S) TestDialWithMixedSeedList(c *C) {
```
niemeyer, line 1020: I don't understand the benefit of all the tests above if we have this one. We might even test the log to ensure the proper error was reported, but this makes it clear that the offending servers never get into the live list. We don't need to test again that mgo behaves correctly when no appropriate servers are available, as we have specific tests checking just that.

session.go
```diff
@@ -294,9 +298,16 @@ type DialInfo struct {
...
 	// Database is the default database name used when the Session.DB method
 	// is called with an empty name, and is also used during the intial
-	// authenticatoin if Source is unset.
+	// authentication if Source is unset.
 	Database string
 
+	// ReplicaSetName defines the name of the replica set to use for cluster
+	// discovery and monitoring. If specified, at least one of the seed servers
+	// must be a member of the ReplicaSetName replica set - non-members are
+	// ignored. If unspecified, the intended connection is assumed to be either
```
niemeyer, line 307: That's not true. If unspecified the driver can still speak to a replica set. Again, no need to fix.. I'm merging and doing the needed changes.

niemeyer:
This was merged into v2-dev, with the trivial changes from e0d7e593b, and will be in the next release.

Thanks for the contribution.

cluster_test.go
deafgoat, line 1020: Yup, sounds good to me. Is this eventually going to be in the v2 branch?

session.go
deafgoat, line 307: You're right, thanks.

deafgoat:
Thanks for taking the time to review!

cluster_test.go
niemeyer, line 1020: Yeah, definitely. v2-dev contains the content that will be the next v2 release.

sundarv85:
@deafgoat @niemeyer I was looking for the ReplicaSetName in the [latest documentation](http://godoc.org/labix.org/v2/mgo#DialInfo) - but I’m unable to find it. Am I looking at the correct location? Is this ReplicaSetName feature now removed..

We have a replica set and only want to communicate to the primary and not every server that is discovered. How can I achieve that from the current mgo package.

deafgoat:
Hi @sundarv85,

You can view the v2 documentation of the mgo driver at http://godoc.org/gopkg.in/mgo.v2 - which also contains documentation on the `ReplicaSetName` feature.

To communicate with just the primary (or seed server(s)) of a replica set), it suffices to set the `Direct` field of the `DialInfo` struct - so no server discovery is performed. See [here](https://github.com/deafgoat/mgo/blob/84603f4983eeaf751409ad2620640060714f7a4d/cluster_test.go#L1135) for an example.

The `ReplicaSetName` name feature simply ensures that if `Direct` is unset and you dial MongoDB with seed servers from different replica sets, only those servers that are part of the the specified replica set are communicated with. This can be achieved either by using the `DialInfo` struct or a [connection URI](https://github.com/deafgoat/mgo/blob/84603f4983eeaf751409ad2620640060714f7a4d/cluster_test.go#L945).

PR #60, opened 2015-01-05
Bulk inserts now respect the maxMessageSize by splitting up the inserts ...

jtsylve:
...into multiple messages if needed

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -1,5 +1,13 @@
 package mgo
 
+import "gopkg.in/mgo.v2/bson"
+
+const (
+	// maxMessageSize is the maximum size of a single message
+	// that is supported by Mongo
+	maxMessageSize = 48000000
+)
+
 // Bulk represents an operation that can be prepared with several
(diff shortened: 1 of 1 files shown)


jtsylve:
I came across this issue on a project that I'm working on.  If you try to bulk insert a large amount of data at once, the insertion will fail because Mongo has a maximum message size limit.  In this pull request, if the maximum message size is exceeded, mgo will split up the insert into multiple smaller inserts.

niemeyer:
Sorry, but this does not look okay. It's changing the Insert API from being a local operation into running remote commands arbitrarily. We can fix this issue in better ways by being smart at execution time, but I'd prefer to wait to integrate such smartness once the bulk API is completed, integrating recent bulk features, so that we do the logic correctly for both cases.

Can you please open an issue about it so we can track the work?

jtsylve:
This makes sense.  It actually occurred to me after submitting this that I
was adding a side effect to the Insert functionality. I'll open an issue.

PR #72, opened 2015-01-31
Adding option to disable IndexCaching as a DialInfo option.

tdubourg:
The driver currently caches requests made to EnsureIndex (and EnsureIndexKey as a result).

Depending on the infra, this could be problematic and cause cache inconsistency if you're already caching that server-side or somewhere else in your client.

I'm adding an option that allows to disable the caching, the default behavior is unchanged. A unitted has been added as well.

### DIFF
--- a/session.go
+++ b/session.go
@@ -56,21 +56,22 @@ const (
 // need to be updated too.
 
 type Session struct {
-	m            sync.RWMutex
-	cluster_     *mongoCluster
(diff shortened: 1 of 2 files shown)


tdubourg:
Hum, up?

niemeyer:
We shouldn't have an option that turns the documentation wrong. The proper way to do this is to have methods that do not do the caching instead. We'll introduce those soon.

PR #74, opened 2015-02-07
Support encoding byte arrays in structs

### LINKED ISSUE #73: Cannot encode byte arrays
I have a struct that looks like this
```go
type Handle struct {
    Hash [32]byte
}
```

When I try to encode it with bson, I get "reflect.Value.Slice: slice of unaddressable array" coming from https://github.com/go-mgo/mgo/blob/v2/bson/encode.go#L368. In that code, if v is not addressable it can be copied to a slice and then added as binary, so it looks like the following (which works):

```go
if v.CanAddr() {
	e.addBinary('\x00', v.Slice(0, v.Len()).Interface().([]byte))
} else {
	slice := reflect.MakeSlice(reflect.SliceOf(v.Type().Elem()), v.Len(), v.Len())
	reflect.Copy(slice, v)
	e.addBinary('\x00', slice.Interface().([]byte))
}
```

I can submit a pull request, but it would help if someone explained where I should add tests for this case in `bson_test.go`.

tchajed:
Fix #73.

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -1249,6 +1249,7 @@ var twoWayCrossItems = []crossTypeItem{
 
 	// arrays
 	{&struct{ V [2]int }{[...]int{1, 2}}, map[string][2]int{"v": [2]int{1, 2}}},
+	{&struct{ V [2]byte }{[...]byte{1, 2}}, map[string][2]byte{"v": [2]byte{1, 2}}},
 
 	// zero time
 	{&struct{ V time.Time }{}, map[string]interface{}{"v": time.Time{}}},
--- a/bson/encode.go
+++ b/bson/encode.go
@@ -365,7 +365,13 @@ func (e *encoder) addElem(name string, v reflect.Value, minSize bool) {
 		et := v.Type().Elem()
 		if et.Kind() == reflect.Uint8 {
 			e.addElemName('\x05', name)
-			e.addBinary('\x00', v.Slice(0, v.Len()).Interface().([]byte))
+			if v.CanAddr() {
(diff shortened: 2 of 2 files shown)


niemeyer:
I'm still on the fence about this one. It feels surprisingly costly for such a trivial operation.

tchajed:
The problem is that if CanAddr() returns false, you can't address the array's memory. This makes it tricky to pass the bytes in the array to the encoder; since Go doesn't have generics, you can't write a method that takes an arbitrary array, it has to take a slice, and there's no way to get a zero-copy slice in this case.

One way around the copy is to instead pass the bytes individually to the encoder. This will have a high reflection overhead, since you'll have to call `(v Value) Index()` and convert each `reflect.Value` to a byte.

I agree that a copy seems unnecessary, but it's better than one reflection conversion per byte or leaving the bug there.

niemeyer:
The trade-off doesn't seem so clear to me as it is to you. You see it as a bug, and I see it as a feature explicitly left out for lack of a reasonable way to handle it. Requiring 3000 allocations to encode 1000 values which contain 3 arrays with a couple of integers each sounds bad, even ignoring the fact that they need to be copied. The impact is disproportional to the expectation of what such an operation would do.

I'm not rulling it out yet, but it's not so straightforward a choice to me.

tchajed:
I checked what encoding/json does and it's actually the other option I suggested (access each element as a `reflect.Value` and then get the element). This will be slower but won't cause another memory allocation.

niemeyer:
Okay, I'd be happier about that implementation as well.

tchajed:
Cool, implemented that in pull #110.

PR #75, opened 2015-02-09
Run gofmt and goimports over code

tchajed:
It's fairly standard for Go projects to be formatted according to gofmt.

This pull request also includes the simplifications from `gofmt -s`; the one that shows up is that in a slice or map literal containing literals, the type in the element literals can be omitted. `map[string][]int{"array": []int{5, 9}` to `map[string][]int{"array": {5, 9}` change is a good example, where the `[]int` type is not necessary.

### DIFF
--- a/auth_test.go
+++ b/auth_test.go
@@ -262,7 +262,7 @@ func (s *S) TestAuthUpsertUserOtherDBRoles(c *C) {
 	ruser := &mgo.User{
 		Username:     "myruser",
 		Password:     "mypass",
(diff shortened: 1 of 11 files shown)


niemeyer:
We very much follow go fmt rules, as you can observe by how small the patch set is. Sometimes things do go in without following the rules, though (merges, etc).

I'm happy to apply your change set, either way, but not on v2. Any changes should go into v2-unstable.

PR #79, opened 2015-03-01
Adding MarshalText and UnmarshalText methods to ObjectId

### LINKED ISSUE #78: BSON ID Values And XML/YAML Marshal/Unmarshal
Hi,

When marshaling/unmarshaling to/from XML I noticed that BSON ID values don't render as recognizable strings.  Instead they reader as byte values.

Is there a reason we wouldn't want to include this support with the following methods below in the `bson.go` file?  They are the result of a conversation between @klauspost and @niemeyer here: https://groups.google.com/forum/#!topic/mgo-users/uJqqe_tqCuE

```Go
// MarshalText returns the ObjectId in hex format.
func (id *ObjectId) MarshalText() (text []byte, err error) {
	return []byte(id.Hex()), nil
}

// MarshalText reads the hex value of a marshalled field, and sets the current ID to that value.
// If the value is not a valid hex value, an error is returned
func (id *ObjectId) UnmarshalText(text []byte) error {
	if !IsObjectIdHex(string(text)) {
		return errors.New("invalid bson.ObjectId value")
	}
	*id = ObjectIdHex(string(text))
	return nil
}
```

Sorry if this has already been discussed or resolved.  Thanks in advance!

jackspirou:
This pull request fixes #78.

It is being added to the `v2` branch, but i'm not sure if it should be included into `v2-unstable` as well.

Thanks for the help.

Best, 
- Jack

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -34,8 +34,8 @@ import (
 	"fmt"
 	"sync"
 
-	"gopkg.in/mgo.v2-unstable/bson"
-	"gopkg.in/mgo.v2-unstable/internal/scram"
+	"gopkg.in/mgo.v2/bson"
+	"gopkg.in/mgo.v2/internal/scram"
 )
 
 type authCmd struct {
--- a/auth_test.go
+++ b/auth_test.go
@@ -39,7 +39,7 @@ import (
 	"time"
 
 	. "gopkg.in/check.v1"
-	"gopkg.in/mgo.v2-unstable"
+	"gopkg.in/mgo.v2"
 )
 
 func (s *S) TestAuthLoginDatabase(c *C) {
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -283,6 +283,29 @@ func (id *ObjectId) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// MarshalText turns bson.ObjectId into an encoding.TextMarshaler.
+func (id ObjectId) MarshalText() ([]byte, error) {
+	return []byte(fmt.Sprintf("%x", string(id))), nil
+}
+
+// UnmarshalText turns *bson.ObjectId into an encoding.TextUnmarshaler.
+func (id *ObjectId) UnmarshalText(data []byte) error {
+	if len(data) == 1 && data[0] == ' ' || len(data) == 0 {
+		*id = ""
+		return nil
+	}
+	if len(data) != 24 {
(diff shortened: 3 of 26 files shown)


niemeyer:
There are a few conditions that we use in the json unmarshaler logic which seem appropriate for the text marshaler as well. Can you please have a look and let me know what you think?

Also, I believe the marshaler should take the receiver by value, and not as a pointer. It is a string, so there's no benefit in receiving it as a pointer in this case.

jackspirou:
Both points make sense.  I attempted to addressed them below.
1. I took a look at the `UnmarshalJSON` logic and tried to apply it where it made sense.  Please take a look and let me know if I missed a particular case you would like to include, I won't be offended :)
2. I have updated the marshaler to take the receiver by value and not as a pointer.  It is also worth pointing out that the `MarshalJSON` method takes the receiver by value as well.

niemeyer:
Thanks for these changes. The only thing to fix here that I could notice is that strange special casing of the a string which is a single empty space. If we want to handle empty spaces as an unset id, any number of spaces should have the same effect.

The other bit is that this is missing tests. Otherwise, it looks good.

niemeyer:
As a hint, the check you want there is likely something along these lines:

```
if len(bytes.TrimSpace(data)) == 0 { ... }
```

niemeyer:
One more detail: can you please rebase this against v2-unstable?

jackspirou:
closing this pull request to start fresh

jackspirou:
@niemeyer I think I am ready to open a new pull request with these changes: https://github.com/go-mgo/mgo/compare/v2...jackspirou:v2

Do you want me to target the `v2` branch or the `v2-unstable` branch?  Sorry about closing this pull request, I wasn't sure if you wanted me to `git rebase v2-unstable` or target the `v2-unstable` branch to submit the changes against.  I thought I would start clean since `git rebase v2-unstable` caused conflicts for me.

Let me know how you would like me to submit this pull request.  Thanks for your patience.

rodriguise:
Was this ever merged? Will it make it into the stable codebase?

jackspirou:
@rodriguise this was never merged, but I did just submit the changes to the unstable codebase at https://github.com/go-mgo/mgo/pull/212. cc @niemeyer.

niemeyer:
Thanks, I'll have look.

PR #83, opened 2015-03-28
Adjust stats for master connection only once.

dmage:
Value of SlaveConns may become negative, as each topology synchronization may reduce it by 1.


niemeyer:
Thanks!

niemeyer:
Oops.. this patch sneaks in a serious breakage in the cluster synchronization logic which is completely unrelated to fixing the statistics. I'll let you guess which line does that. :-)

niemeyer:
Also note that existent tests prevented this from being released, which probably means they were not run.

dmage:
Sorry, this is true, I didn't notice tests. I added a test for this case in #109.

PR #85, opened 2015-04-03
Capitalize ParseURL to make public.

rgbkrk:
To prevent us from writing boilerplate code for parsing a Mongo URL (or having to provide the options directly), it would be great to have the ParseURL function be public.

Related issue: https://github.com/cloudpipe/cloudpipe/pull/69

### DIFF
--- a/session.go
+++ b/session.go
@@ -218,7 +218,7 @@ func Dial(url string) (*Session, error) {
 //
 // See SetSyncTimeout for customizing the timeout for the session.
 func DialWithTimeout(url string, timeout time.Duration) (*Session, error) {
-	uinfo, err := parseURL(url)
+	uinfo, err := ParseURL(url)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


rgbkrk:
Apparently others have wanted this, because there's a [fork of just the parseURL function](https://github.com/jmcvetta/mgourl). It doesn't have tests and I'd rather see it as part of this package since you update it regularly.

niemeyer:
Sounds good. It takes slightly more than capitalizing the function, but the idea is sound.

rgbkrk:
Consider this my humble issue if you prefer, I'm happy if there's more you'd like to see. :wink:

niemeyer:
Thanks, this is definitely an often-requested feature, and I'll make sure to get it in place for you. #97 was a bit closer to the end result, but still a bit off in that we need the result to be a pointer, which is what the public API expects. I'll get something in and let you know via this PR.

Issue #84, opened 2015-04-01
Support SSL


rgbkrk:
Love the library, thanks for it. I'm hoping to use it with a Mongo server that uses SSL, passing a mongodb URL like so:

```
mongodb://USERNAME:PASSWORD@iad-mongos2.objectrocket.com:PORT/dbname/?ssl=true
```

This is based on the `url.ssl` section of the [connection string documentation](http://docs.mongodb.org/manual/reference/connection-string/#uri.ssl).

niemeyer:
To connect to an SSL-enabled MongoDB, just define the DialServer 
function in [mgo.DialInfo](http://gopkg.in/mgo.v2#DialInfo), and make use of the tls.Dial [2] to 
perform the connection. 

Please let me know if you need any further details.

rgbkrk:
That's fine, and what we've done so far. It would be nice if we could use the mongo URL parser so we don't have to construct the entire DialInfo object.

```
dialInfo := &mgo.DialInfo{
	Addrs:    []string{c.Settings.MongoURL},
	Database: c.Settings.MongoDB,
	Username: c.Settings.MongoUser,
	Password: c.Settings.MongoPassword,
	DialServer: func(addr *mgo.ServerAddr) (net.Conn, error) {
		return tls.Dial("tcp", addr.String(), &tls.Config{})
	},
	Timeout: time.Second * 10,
}
session, err := mgo.DialWithInfo(dialInfo)
```

mhurne:
I understand the motivation/reasoning behind the approach described above, but I wonder if `ParseURL` could at least parse the `ssl` option and let the caller know what its value was? Perhaps something like:

```
func ParseURL(url string) (*DialInfo, bool, error) {
```

Where the bool is true if the URI included `ssl=true` and `false` otherwise?

When writing a general-purpose client that will be configured using a Mongo URI, the current API requires extracting the `ssl` param from the URI before passing it to `ParseURL` or writing alternative URI parsing logic rather than using `ParseURL` at all. In particular, this is a challenge I've faced in my attempt to tackle https://github.com/hashicorp/vault/issues/362

niemeyer:
Yes, that would be nice. We can error out inside DialWithInfo itself if we receive SSL true, pointing people to some useful documentation.

mhurne:
Great - is this something you have time to pursue? Or would you prefer I take a stab at it and submit a pull request?

niemeyer:
[...]

PR #86, opened 2015-04-03
Parse the SSL flag automatically

rgbkrk:
This uses the host's root CA set by default, parsing out [the `ssl` option from the connection string](http://docs.mongodb.org/manual/reference/connection-string/#uri.ssl).

### DIFF
--- a/session.go
+++ b/session.go
@@ -28,6 +28,7 @@ package mgo
 
 import (
 	"crypto/md5"
+	"crypto/tls"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rgbkrk:
Fixes #84.

niemeyer:
We can include stock support for the ssl flag somewhere in the driver, but not in the main package like this. This would force the inclusion of crypto/tls and all its non-trivial dependencies on every application building the driver. Instead, I'd prefer to have an auxiliary package that may be explicitly imported to introduce the SSL support.

niemeyer:
By the way, thanks for the contribution. I appreciate the interest, and would be glad to figure out with you a proper way to do this.

rgbkrk:
Great, thanks for the response @niemeyer.

Issue #87, opened 2015-04-04
The dollar ($) prefixed field '$pullAll' in '$pullAll' is not valid for storage


smanji:
I'm using MongoDB 2.6.6 and using the mgo/txn package to do an update on one collection and an insert in another.

Executing this gives the following error: The dollar ($) prefixed field '$pullAll' in '$pullAll' is not valid for storage

Per the following thread, "$ is not allowed at the beginning of a field name, since in this position it is reserved to indicate query operators like $or, $addToSet, $, etc, and allowing it as part of field names would make the query language ambiguous":
https://groups.google.com/forum/#!topic/mongodb-user/uViuJsl4jP8

Can this package be updated to support Mongo 2.6?

Thanks!

smanji:
I think I fixed this issue with the following pull request: https://github.com/go-mgo/mgo/pull/88

Please review and if you agree, merge it into the original and we can close this issue.

Thanks!
Sadiq.

niemeyer:
The txn package works in MongoDB 2.4 and 2.6, and probably earlier ones. The error you describe doesn't make sense to me, as it reports an attempt to store a $ field, which $pullAll should of course not be as this is an update operator.  The pull request also does not make sense, as it is turning such an update operator into a normal field, thus breaking the package (there is a test suite you can run to test your changes).

Can you please provide a tiny and self-contained test case demonstrating the failure you observed?

smanji:
Hi Gustavo,

Thanks for your response. I'm running mongodb 2.6.6 on Mac OSX.

Here's the tiniest test case I could create to reproduce my issue, it's the second transaction that gives the error -- hopefully you can point out what I'm doing wrong.

Incidentally, my fix in pull request #88 resolves this issue for me... sorry for not testing it with your test suite though!

```Go
package main

import (
	"fmt"
	"gopkg.in/mgo.v2"
	"gopkg.in/mgo.v2/bson"
	"gopkg.in/mgo.v2/txn"
	"time"
)

type ParentStruct struct {
	Id          bson.ObjectId `json:"id,omitempty" bson:"_id"`
	UpdatedDate time.Time     `json:"updated_date" bson:"updated_date,omitempty"`
}

type ChildStruct struct {
	Id       bson.ObjectId `json:"id,omitempty" bson:"_id"`
	ParentId bson.ObjectId `json:"parent_id" bson:"parent_id"`
}

func main() {
	mgoSession, _ := mgo.Dial("localhost")
[...]

PR #88, opened 2015-04-06
Fixed "The dollar ($) prefixed field '$pullAll' in '$pullAll' is not valid for storage" issue

smanji:
Fixed issue that prevented ability to process updates to mongodb documents in v2.6 due to 'The dollar ($) prefixed field '$pullAll' in '$pullAll' is not valid for storage' error, and subsequent error with $set.

It's a simple fix (just removed the $ sign in two places) and it seems to be working for my code now.

Please review and merge back into original v2 if you agree with the changes.

Thanks!
Sadiq.

### DIFF
--- a/txn/flusher.go
+++ b/txn/flusher.go
@@ -734,10 +734,10 @@ func (f *flusher) apply(t *transaction, pull map[bson.ObjectId]*transaction) err
 				if d, err = objToDoc(op.Update); err != nil {
 					return err
 				}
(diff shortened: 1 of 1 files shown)


niemeyer:
The txn package works in MongoDB 2.4 and 2.6, and probably earlier ones. The error you describe doesn't make sense to me, as it reports an attempt to store a $ field, which $pullAll should of course not be as this is an update operator.  The pull request also does not make sense, as it is turning such an update operator into a normal field, thus breaking the package (there is a test suite you can run to test your changes).

Please follow up in issue #87.

PR #90, opened 2015-04-06
add support for geoHaystack index

deiwin:
As seen [in the manual](http://docs.mongodb.org/manual/tutorial/build-a-geohaystack-index/), the haystack index uses a property called bucketSize, which isn't currently available through this driver. This commit adds the property.

**Note**: I'm not entirely sure if the tag key is actually needed for the `BucketSize` field (i.e. `"bucketSize,omitempty"` vs `",omitempty"`). The [documentation](http://godoc.org/gopkg.in/mgo.v2/bson#Marshal) seems to indicate that the key should be the lowercased field name anyway. But without the key the added test failed. Although, I didn't spend too long setting up the environment and some other, seemingly random, tests failed as well. So it could be that my envrionment is to blame. But then I saw that `DropDups` also specified the tag key, so I'm not really sure either way.

### DIFF
--- a/session.go
+++ b/session.go
@@ -947,15 +947,16 @@ func (db *Database) RemoveUser(user string) error {
 type indexSpec struct {
 	Name, NS         string
 	Key              bson.D
-	Unique           bool   ",omitempty"
-	DropDups         bool   "dropDups,omitempty"
-	Background       bool   ",omitempty"
-	Sparse           bool   ",omitempty"
-	Bits, Min, Max   int    ",omitempty"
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks for the change. The field name is required because the default is to fully lowercase the field name, or "bucketsize" in this case, which isn't correct.

niemeyer:
By the way, shouldn't the bucket size be an integer value?  Does it make sense to provide fractional bucket sizes?

niemeyer:
Nevermind.. confirmed in the C++ code that it is a double.

PR #96, opened 2015-05-06
Avoid Setter creation if value is nil

roman-mazur:
Check for nil is now done before the code that handles `Setter` interface.
Otherwise Setter instance is created even if actual value is nil, which is quite unexpected.

Fixed scenario is illustrated in `TestUnmarshalNilSetter`.
I had to modify `TestUnmarshalAllItemsWithPtrSetter` in an awkward way though.

After implementing this change I found out that `SetBSON` can return `SetZero`. But I personally do not understand how creating a `Setter` instance can be useful if we unmarshal `nil` value.
Will appreciate if an example is given.
It also makes `SetBSON` method implementation a bit gross IMO since you have to write code like

``` go
if raw.Kind == 0x0A { // Nil value.
  return bson.SetZero
}
```

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -770,8 +770,17 @@ func (s *S) TestUnmarshalAllItemsWithPtrSetter(c *C) {
 				}
 			} else {
 				expected := item.obj.(bson.M)["_"]
-				c.Assert(field, NotNil, Commentf("Pointer not initialized (%#v)", expected))
-				c.Assert(field.received, DeepEquals, expected)
+				if expected == nil {
+					if i == 0 {
+						c.Assert(field, IsNil, Commentf("Nill expected. Got (%#v).", field))
+					} else {
+						// Field is not nil since it's a value.
+						c.Assert(field.received, IsNil, Commentf("Nill expected. Got (%#v).", field.received))
+					}
+				} else {
+					c.Assert(field, NotNil, Commentf("Pointer not initialized (%#v).", expected))
+					c.Assert(field.received, DeepEquals, expected)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


niemeyer:
This patch would prevent people from using YAML's null value to initialize Go values, for no benefit apparently. As you point out in the proposal, it's currently supported and trivial to choose to have a zero value instead of the initialized value, if that's what you want.

roman-mazur:
@niemeyer Could you show a small example how this feature is used? Out of curiosity...

niemeyer:
Can you provide more details about exactly what you are asking for and why?  Will be happy to provide you with an example once you explain what you are trying to achieve and are being unable to.

roman-mazur:
What I want to achieve is omitting an allocation of my type when `nil` is deserialized.
I do not have strong arguments that it impacts performance in any way. I just do not understand why this should happen.

In Go structure

``` go
type Container struct {
  value *MyCustomTypeThatImplementsSetter
}
```

I define `value` as a pointer. This means that `value` can be `nil`. And if `Container` is unmarshaled from bson `{"value":nil}`, I expect the `value` field in `Container` to become `nil`. And it does not make sense to allocate a new instance of `MyCustomTypeThatImplementsSetter` to set `value` to `nil`.

Actually marshalling works in the same way. If you define a method

``` go
func (p *SomeType) SomeMethod() {
}
```

you usually do not expect `p` to be `nil`. And it's what happens when `nil` is serialized with this `bson` package.

AFAIK neither `json.Unmarshaler` nor `json.Marshaler` method will not be invoked in case of `nil`. At least it's what I experience implementing both `SetBSON` and `UnmarshalJSON` (`GetBSON` and `MarshalJSON`) on my type.

Finally, even if we think that allocating `Setter` to unmarshal `nil` is ok, `SetBSON` implementation becomes dirty (IMO).
One of intents to use `bson` package is to avoid dealing with bson format directly. But the only way to understand that underlying value is nil (please correct me if I do not see something) is to write

``` go
if raw.Kind == 0x0A { 
  return bson.SetZero
}
```

And to write this code you have to dive into bson spec and see that `nil` is encoded as `0x0A`.

Just sharing my experience and feelings :)

PR #97, opened 2015-05-08
Expose ParseURL method.

cmars:
Currently, there's no way to determine the options that were given in a mongo connection URL string in your application, other than duplicating the implementation of `parseURL` (which is quite unappealing). It would be nice to be able to use the same parser that `mgo` uses.


niemeyer:
This is a dup of #85.

PR #99, opened 2015-05-11
Check PurgeMissing iterator errors & other minor fixes

mjs:
The main fix here is to have PurgeMissing check for pipeline iteration errors. Not having these meant that PurgeMissing was silently failing in a production recovery situation (due to pipeline result size limits under MongoDB 2.4). A later PR will propose a change to PurgeMissing to avoid the pipeline size limit issue.

The other commits are all tiny fixes to PurgeMissing and TestPurgeMissing.

### DIFF
--- a/txn/txn.go
+++ b/txn/txn.go
@@ -417,6 +417,9 @@ func (r *Runner) PurgeMissing(collections ...string) error {
 				return fmt.Errorf("error purging missing transaction %s: %v", txnId.Hex(), err)
 			}
 		}
+		if err := iter.Close(); err != nil {
+			return fmt.Errorf("transaction queue iteration error for collection %s: %v", collection, err)
+		}
 		colls[collection] = true
 	}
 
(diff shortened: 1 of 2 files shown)


txn/txn.go
```diff
@@ -417,7 +416,9 @@ func (r *Runner) PurgeMissing(collections ...string) error {
 				return fmt.Errorf("error purging missing transaction %s: %v", txnId.Hex(), err)
 			}
 		}
-		colls[collection] = true
+		if err := iter.Close(); err != nil {
+			return fmt.Errorf("transaction queue iteration error for %s: %v", collection, err)
```
niemeyer: "%s" => "collection %s", please.

txn/txn.go
```diff
@@ -442,6 +443,9 @@ func (r *Runner) PurgeMissing(collections ...string) error {
 			return fmt.Errorf("error purging missing transaction %s: %v", txnId.Hex(), err)
 		}
 	}
+	if err := iter.Close(); err != nil {
+		return fmt.Errorf("stash iteration error: %v", err)
```
niemeyer: s/stash/transaction stash/, please.

niemeyer:
Just a couple of trivials. Please ping me when done and I'll merge it.

txn/txn.go
mjs: Ha :) This is what I originally had.

PR #100, opened 2015-05-12
Guard against empty/malformed transaction ids in storage.

cmars:
A panic may result from a transaction document that contains an invalid `Id`. This is not normally expected behavior but could happen if the disk filled up, for example.

My solution is to add a `txn/transaction.valid()` method, which is checked at the top of methods that otherwise assume the `Id` is valid (creating tokens for them makes this assumption).

Its possible that there are better ways to handle these errors. If so, please advise.

Encountered this issue in the wild (LP:#1452251).


niemeyer:
We discussed this concept out of band.

PR #109, opened 2015-05-26
Add test for connections statistics after syncing

dmage:
This is a test case for issue #83. Perhaps there is a better way to run syncServersLoop rather than sleep.

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -1655,3 +1655,24 @@ func (s *S) TestSelectServersWithMongos(c *C) {
 		c.Fatal("Uh?")
 	}
 }
+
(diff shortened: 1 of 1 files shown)


niemeyer:
It doesn't look like this test is doing much to test that issue. We have quite a lot of tests in `cluster_test.go` verifying that the values are basically right.

dmage:
Ok, let's wait for regression.

niemeyer:
Sounds good. The test as it is is hugely dependent on subtle internal details either way. A regression that breaks this logic could trivially turn the test irrelevant in one go.

PR #110, opened 2015-05-26
Support encoding byte arrays in structs

### LINKED ISSUE #73: Cannot encode byte arrays

tchajed:
Indexes each element of the array rather than allocating a slice to
copy the array into as in #74.

Fixes #73.

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -1268,6 +1268,7 @@ var twoWayCrossItems = []crossTypeItem{
 
 	// arrays
 	{&struct{ V [2]int }{[...]int{1, 2}}, map[string][2]int{"v": [2]int{1, 2}}},
+	{&struct{ V [2]byte }{[...]byte{1, 2}}, map[string][2]byte{"v": [2]byte{1, 2}}},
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks!

PR #113, opened 2015-06-03
Replace bad not found check in Update.

tcard:
`Update` was assuming that, if no document was updated, the selector query wasn't matching anything, thus returning `ErrNotFound`. That's wrong, because it can happen that the selector matches but then the new updated value is jus the same as the old one, in which case `nModified` is 0 even if the query is successful.

Now `LastError` has a `Found` method that returns true if `result.N` is greater than zero. `Update` checks that method instead of `UpdatedExisting`.

I haven't added a test nor run the existing ones because apparently they need a quite complex previous setup in your machine. Sorry for that.

### DIFF
--- a/session.go
+++ b/session.go
@@ -2174,6 +2174,10 @@ func (err *LastError) Error() string {
 	return err.Err
 }
 
+func (err *LastError) Found() bool {
+	return err.N > 0
+}
+
 type queryError struct {
 	Err           string "$err"
 	ErrMsg        string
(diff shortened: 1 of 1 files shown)


niemeyer:
Thanks for the contribution. Let's please avoid adding a new method, as it's unrelated to the bug being fixed and we don't need to agree on it right now (LastError.Found is misleading, for example). Then, we do need tests for it. You can start servers in your machine by running "make startdb" before running the test suite. If you don't want to do this, please just open an issue including a snippet that demonstrates the bug and I'll fix it locally.

niemeyer:
Also, please mention which server versions you are testing this under.

niemeyer:
Sorry, this cannot wait. I've fixed it and will tag a new release shortly.

niemeyer:
This is out.

tcard:
I wasn't able to make the changes in time. Thanks for taking care of it.

PR #122, opened 2015-06-16
add Extra field to CollectionInfo

reusee:
This is to allow setting options which provided only by alternative implementations of MongoDB, mainly TokuMX: http://docs.tokutek.com/tokumx/tokumx-collection-index-options.html


niemeyer:
Supporting such a wildcard option would mean preventing extending that structure to support further options in the future in a backwards compatible way, as we've done multiple times over the history of the driver (and have just done so again days ago). It might be fine to support actual options, though.

PR #126, opened 2015-06-22
Deduplicate & optimize decode code with reflection

pascaldekloe:
The reflection overhead is compensated with fewer memory allocations.


niemeyer:
This is introducing the use of FieldByName twice in the fast path of a loop. We don't have a single call of that method so far, and I'm not quite convinced doing so is a good idea. If you have a benchmark showing this has a positive impact rather than a negative one in relevant cases, please feel free to reopen the ticket.

PR #128, opened 2015-06-29
less -> fewer

rgarcia:
Small thing I noticed reading the code :smile:


niemeyer:
Thanks, but I believe this is grammatically sound, and also irrelevevant if you understood the intended meaning.

PR #130, opened 2015-07-02
Add functions to expose cursorID and batch size of iterators

dmliao:
Hi, I'm working on a mongo proxy server that uses mgo to communicate with MongoDB, and needed the ability to be able to separate the find and getMore operations into their distinct operations (rather than how Iterators automatically call getMore once the current batch is exhausted). I didn't see any way of manually emulating a getMore operation besides calling `NewIter` on a collection with an existing cursor ID, so the changes are assuming that method.

The first change is a function that exposes an Iterator's cursor ID, as currently you can create a new Iterator with an existing cursor ID, but no way to retrieve the cursor ID from an existing Iterator object. The second change is a function to set a batch size per Iterator, as creating an Iterator with NewIter doesn't let the user set the batch size, unlike when an Iterator is created with a Query or a Pipe.

### DIFF
--- a/session.go
+++ b/session.go
@@ -3356,6 +3356,22 @@ func (iter *Iter) All(result interface{}) error {
 	return iter.Close()
 }
 
+// CursorID retrieves the cursor ID associated with the iterator.
+//
+// The CursorID is only set once the first document is retrieved from the result
+// set.
+func (iter *Iter) CursorID() int64 {
+	return iter.op.cursorId
+}
+
+// SetBatch sets the batch size used for this iterator when fetching documents from the
+// database.
+func (iter *Iter) SetBatch(n int) {
+	iter.m.Lock()
+	iter.op.limit = int32(n)
+	iter.m.Unlock()
+}
+
 // All works like Iter.All.
 func (q *Query) All(result interface{}) error {
 	return q.Iter().All(result)


niemeyer:
Hi Diana,

I'm afraid your use case is beyond what mgo is meant to offer. For emulating a proxy, you'll find many shortcomings in this driver, because its goal is to abstract away the wire protocol and command mechanism of the server into something that is convenient for developers to use when building their applications, not to be a transparent mechanism that a proxy server can use to fully introspect actions and dumbly replay incoming requests into the server.

dmliao:
Hi Gustavo,

Thank you for your insight, and I appreciate your concerns with using the driver in this way. I'm not building a dumb proxy (one that copies bytes from client to server), but rather one that also abstracts away parts of the wire protocol and provides extra functionality. I can explain the project in more detail, but I don't think it's necessary for these changes.

Can we include these changes and discuss future changes as needed?

niemeyer:
We can't merge this unless we find a non-artificial reason why a normal application would use these details. As far as I can see, it makes no sense to change batch size during iteration, and nobody ever asked me to expose the cursor ID, since the driver is abstracting iterations in a very convenient way, precisely so that they don't need to care about these internals.

dmliao:
Mgo is also currently the only well-maintained MongoDB driver for Go, so without exposing some of those internals the alternative I (and other developers who do care about those details) have would be to write an entirely new driver, which seems redundant when mgo already has most of the feature set we'd be looking for.

I agree with you that batch size shouldn't change during iteration; the original intention was for there should be a way to set the batch size when creating a new Iterator. It should probably be set as an argument for the NewIter function, though that would introduce a breaking change for any application using the NewIter function currently. I can modify the pull request as needed if that is preferred.

I still think that it's important that batch size be settable when creating a new Iterator with NewIter. If developers are using the NewIter function, then they're already working with the internals of MongoDB by providing a firstBatch and a cursorID, and not allowing them to set a batch size makes the implementation seem incomplete.

Although the driver's focus is on abstracting away internals for simplifying development, I think the internals can be exposed as long as it doesn't detract from the design or complicate the development process, and we already do so a little bit via the NewIter function. Exposing CursorID or being able to set a batch size for new iterators won't make working with mgo any more difficult for app developers, as it doesn't change the existing abstractions.

niemeyer:
The fundamental question is why do you want to create a new iterator? This is the sort of task the driver itself should be doing for you. There's a very special reason why NewIter exists, and one that took some significant thinking time. I wasn't keen on exposing it at all to begin with, and it's not an excuse to open the gates to functionality that is added without justification.

Regarding creating a new driver, if your use case is incompatible with the goals of this driver, I see no problem with creating another independent driver that is tailored for you own use case. This driver is not meant to be a proxy, and I'm not keen on adding functionality that is completely irrelevant for every other use case but the proxy one.

dmliao:
The functionality does have other use cases, though. Having access to the cursorID is important to correlate logs between the client application and the MongoDB server, which would be useful for diagnostics. The MongoDB server's logs use the cursorID as an identifier for different queries, and have additional information that isn't sent to the client, including the amount of time it took the server to process that request (useful for diagnosing slow queries), among other things.

niemeyer:
I've never had someone mention that use case, and we can easily add cursor id to the information we log if that becomes necessary. That's what I meant by non-artificial. I'm interested in what is your use case for this, not the potential use case of a theoretical user. Once more, what is your use case for creating iterators manually, and why do you need a cursor id?

niemeyer:
Closing for lack of feedback.

PR #135, opened 2015-07-20
MGO-64: GridFS completeWrite doesn't create unique index

deafgoat:


niemeyer:
Thank you!

niemeyer:
Please note I've tweaked the logic so the file isn't removed if index creation fails.

PR #137, opened 2015-07-23
Minor documentation fix

rychipman:
Just a little bit of missing documentation I came across while reading through the connection-string-parsing logic.


niemeyer:
Thanks! Looks like it's not properly indented, but I'll fix that locally.

PR #138, opened 2015-07-25
Session.IsDup for *bulkError by underlying error

ek220:
Session.IsDup for *bulkError by recursion using underlying err

### DIFF
--- a/bulk_test.go
+++ b/bulk_test.go
@@ -61,6 +61,7 @@ func (s *S) TestBulkInsertError(c *C) {
 	bulk.Insert(M{"_id": 1}, M{"_id": 2}, M{"_id": 2}, M{"_id": 3})
 	_, err = bulk.Run()
 	c.Assert(err, ErrorMatches, ".*duplicate key.*")
+	c.Assert(mgo.IsDup(err), Equals, true)
 
 	type doc struct {
 		N int `_id`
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks, sounds reasonable.

Can we please have a test for this?

ek220:
Added asserts to bulk tests
Testing the IsDup directly is impossible due to bulkError is private

-- Eugeny

niemeyer:
Okay, that sounds fine. There's a small question here, though. What is the semantics of IsDup when one of the errors is a dup but the other is not?

ek220:
At the moment bulkError is container for single error.
For single error IsDup is applicable. 

The possible solution to provide array contains an error document for each write operation that errors.
http://docs.mongodb.org/manual/reference/method/BulkWriteResult/#BulkWriteResult

niemeyer:
The reason bulkError is not exposed is precisely because it is not yet representing everything it needs to for bulk scenarios. If the idea was to have a single error always, it wouldn't make sense to wrap it in such a way.

niemeyer:
Maybe we need something along the lines of `BulkError.Has` or `BulkError.Contains`, which takes a predicate function of the `IsDup` style. Another option would be to say that such standard predicates return true if any one of the errors match, which doesn't smell so great, but maybe it's the useful thing to do. We need to think through what would be the correct way to handle such multi-error cases.. what will we want to do when we have one of these at hand?

niemeyer:
After sleeping on this, we should make IsDup return true when all errors in a bulk update are dup errors, which means it would be fine to make IsDup return true when there's on a single error and that error is a dup error. That seems like the most useful and safe approach.

ek220:
Need more investigations

niemeyer:
Why? I was tempted to adopt your suggestion, for the reason exposed above.

PR #141, opened 2015-07-27
bson: reading json struct tag if bson tag is missing

mcuadros:
This is a very small PR but for sure very controversial, the goal is try to avoid add more tags to the fields, reading the `json` struct tag if the `bson` one is missing. 

I am pretty sure that many projects using mgo and having also json serialzable structs, are like: 

``` go
type Person struct {
    Name              string `bson:",omitempty" json:",omitempty"`
    FullName          string `bson:"full_name,omitempty" json:"full_name,omitempty"`
    Description       string `bson:",omitempty" json:",omitempty"`
}
```

This is really annoying because in many cases is expected to have the same snake cases names in a json and also in the mongo document, also could be a source of bugs due to the requirement of the duplicate tags.

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -600,6 +600,9 @@ func getStructInfo(st reflect.Type) (*structInfo, error) {
 		info := fieldInfo{Num: i}
 
 		tag := field.Tag.Get("bson")
+		if tag == "" {
+			tag = field.Tag.Get("json")
+		}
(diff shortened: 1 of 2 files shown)


niemeyer:
Different packages really have different fields with different potential attributes and meaning. We really can't promise full compatibility with json there, and it's also questionable whether this is always what's desired.

PR #147, opened 2015-08-10
Add primitive decimal support to mgo

raymondjacobson:
I believe/hope these changes now should reflect our discussion.

I had to add two fields to the Decimal struct though:
type Decimal struct {
    finite bool
    precision int
    rational big.Rat
}

The finite boolean is used to determine whether the Decimal is infinite or finite. big.Rat doesn't have a notion of infinite numbers even though MongoDB's Decimal128 (IEEE 754) does. An infinite Decimal stores -1 in big.Rat for negative infinity and +1 in big.Rat for positive infinity along with the boolean flipped.

The precision int is used to store the post-decimal point precision of the value. The reason this is necessary is when encoding a Decimal as BSON, we must invoke the big.Rat.FloatString(prec int) method, which needs a precision (big.Rat defines precision as digits post-decimal point). When a decimal is sent from the server as BSON, we calculate its post-decimal place precision and store that in Decimal.precision. When encoding that decimal back to BSON, we use that precision for Rat.FloatString, which makes round tripping function properly.

When parsing out a precision from a decimal string, if the string is something like "2E-25" where the exponent signals the precision, we assume a max precision of 34. This is somewhat consistent with the notion that big.Rat's method of converting to a string, which never outputs scientific notation because a Rational has no notion of precision. Constructing a big.Rat with "2E-50" has more than 34 digits of "precision" (even though it technically is just 1 digit of precision) and round tripping these values, unfortunately, doesn't work perfectly (outside the range of valid MongoDB 34 digit precision decimals).

More precisely said: In mgo, when working with decimals (as they are big.Rats), there is no notion of cohorts and all values written with scientific notation are extend to regular floating point notation with added precision (since they are actually rational numbers, not decimals). Fixed point functionality can instead be achieved with using precision arguments such as in big.Rat.FloatString().

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -402,6 +402,13 @@ type DBPointer struct {
 	Id        ObjectId
 }
 
+// Uint64x2 is a struct that represents a 128 bit uint and is used
+// for marshaling, parsing, and testing the BSON decimal type
+type Uint64x2 struct {
+	Low uint64
+	High uint64
+}
+
 const initialBufferSize = 64
 
 func handleErr(err *error) {
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -40,6 +40,7 @@ import (
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/mgo.v2-unstable/bson"
+	"gopkg.in/mgo.v2-unstable/decimal"
 	"gopkg.in/yaml.v2"
 )
 
@@ -123,6 +124,8 @@ func (s *S) TestUnmarshalSampleItems(c *C) {
 // length and last \x00 from the document. wrapInDoc() computes them.
 // Note that all of them should be supported as two-way conversions.
 
+var dcml, _ = decimal.Parse("0.1")
+
 var allItems = []testItemType{
 	{bson.M{},
 		""},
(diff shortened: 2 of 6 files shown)


niemeyer:
Thanks for the follow up, Raymond. I'll add a few initial comments here.
1. The +/-infinity approach sounds fine.
2. The precision details are not clear to me. Why would 2E-50 have more precision than supported by MongoDB, when both the significand and the exponent are well within the range of the decimal128 type? Rationals are arbitrary-precision, and MongoDB is not, but both should be able to represent 2E-50 fine.
3. Rather than storing a precision within the Decimal, we should instead compute the precision based on the actual rational value (the divisor, more precisely). This should make it not only roundtrip correctly, but work appropriately once operations are performed with multiple decimal values as well. It also removes the need for manual precision methods and parameters.
4. As we talked, the String method is a strong convention in Go, and needs to have the specific prototype we discussed, without arguments. Otherwise it's not implementing the proper interface and will not work properly in contexts where it should.
5. As we agreed, none of the extra methods and types should be exposed in the bson package. The external interface should remain untouched. For testing, you can create public functions within export_test.go that just call the private ones, and then use these in tests. That will prevent these functions from actually shipping in the package outside of testing.
6. Let's please not create dummy types such as uint32x4 and Uint64x2... Instead, we can use the real underlying types ([4]uint32 and [2]uint64).

Please let me know if you have any questions.

raymondjacobson:
1, 4, 5, 6 -- cool (fixed).

2,3 -- Wow. yeah. Cool.

PR #148, opened 2015-08-11
Add primitive decimal support to mgo

raymondjacobson:
Fixed the decimal precision stuff & addressed your other comments.

Quick question though -- when users enter in decimal.Parse("1/3"), they'll get a precision of 1. Should we just disallow users from Parsing from a string with a "/" in it?

Thanks so much for the review comments!


niemeyer:
Alternative implementation introduced in https://github.com/go-mgo/mgo/commit/14a4475d4d32049500ac89e1499c46be62bf7bdf.

PR #151, opened 2015-08-17
make the mode type public

gabrielrussell:
Having the type "mode" of the constants Eventual, Monotonic, and Strong be private means that you can't create variables of that type, or store the results of Mode() in a data structure.


niemeyer:
Looks good, thanks.

PR #152, opened 2015-08-19
Add EnsureTTLIndex that unconditionally populates ExpireAfter

### LINKED ISSUE #103: Unable to create an expire at TTL index
Trying to create an index with `ExpireAfter: 0` results in the `expireAfterSeconds` option not being applied.

Relevant documentation: http://docs.mongodb.org/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time

My guess is that the `omitempty` bson tag option is preventing the `0` value from being sent to mongo when creating the index. There doesn't seem to be an easy way of supporting this without changing the struct/API in some way?

tmc:
This enables creating TTL indexes with expireAfterSeconds set to 0.

Fixes #103

### DIFF
--- a/session.go
+++ b/session.go
@@ -971,6 +971,11 @@ type indexSpec struct {
 	LanguageOverride string  "language_override,omitempty"
 }
 
+type ttlIndexSpec struct {
+	indexSpec
+	ExpireAfter int "expireAfterSeconds"
+}
+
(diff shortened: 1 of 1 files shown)


niemeyer:
Please see conversation in #103. I'd suggest agreeing on something before spending time on code.

Issue #170, opened 2015-09-28
Need the BatchData and CursorID exposed for our user case


cscs54cs:
Hi,

We are trying do something with `Iter`. Here is our user case:

1. We want to do this on a direct session to a secondary in `Eventual` mode. We make our session using `connect=direct`.
2. We need the `CursorID` and the initial batch.
3. We got these two things and just throw the iter away. We don't close it because we're going to do that separately.
4. We come back at some time, use a session against the same secondary we created the cursor on, we want to continue iterating on that cursor.
5. Repeat the step 4 until we iter all the data for the query.
6. Finally Close the cursor and release the state on the mongo server side.

We have something hacky and are trying to clean it up, but wanted to start a conversation before we get too deep into it. At a high level our approach is to expose a new method on the `Iter` to get the `CursorID` and the first batch, followed by changes to the `NewIter` code path to support using a provided session in `Eventual` mode with a direct connection to a secondary.

Would love to hear your thoughts and see if you have suggestions on how we should solve this.(We're working on putting up our work so far.)

niemeyer:
Hi, and thanks for reaching out.

My initial suggestion would be to keep the iterator around while you need it, instead of trying to poke into the underlying state.

As a side note, why are you using the eventual mode with a direct connection? The point of the direct connection is to pick a defined server, even if it is in a replica set, and the point of the eventual connection is to balance every request across servers. Using both together doesn't seem to make sense.

daaku:
Thanks for the quick response!

> My initial suggestion would be to keep the iterator around while you need it, instead of trying to poke into the underlying state.

We can't (or rather don't) want to do this because we want to avoid our own server maintaining state about the cursor. We want to have mongo maintain the state since it does so anyways.

> As a side note, why are you using the eventual mode with a direct connection? The point of the direct connection is to pick a defined server, even if it is in a replica set, and the point of the eventual connection is to balance every request across servers. Using both together doesn't seem to make sense.

Maybe we're misunderstanding it, but isn't it required to do so if making queries against a secondary? That is, it's equivalent to a `rs.slaveOk()` call in the mongo shell.

In some ways it would probably be easiest for us to do what we're doing by making our own `OP_QUERY` and `OP_GET_MORE` calls and process the data in the `OP_REPLY`s. But our thinking right now is that there are a lot of edge cases to handle, and since mgo already handles them inside the `Iter` implementation we want to leverage it rather than work around it.

niemeyer:
[...]

PR #177, opened 2015-10-09
add support and test for cursor first batch and cursor get more method.

cscs54cs:
Hi, we took a stab at the feature request we described in #170. It exposes 2 new methods to create a new cursor and continue iterating on the cursor. They're definitely plumbing methods and should have some clear docs saying "dont do this unless you know what you're doing". We're using this along with `connect=direct` and `Secondary` mode. We're still not in production with this but wanted to share it early to get your feedback.

For now they're in separate files but that's just so we can keep rebasing with ease. We can move the code to the appropriate files on your guidance.

### DIFF
--- a/cursor_batch.go
+++ b/cursor_batch.go
@@ -1,0 +1,47 @@
+package mgo
+
+import "gopkg.in/mgo.v2/bson"
+
+// CursorGetMore gets a subsequent batch of documents for a given cursorID.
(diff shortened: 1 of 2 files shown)


niemeyer:
This is useful for me to understand what sort of feature you need, but the API is not quite there. It's too close to the internal wire protocol implementation instead of the user-oriented semantics that the driver offers today.

PR #180, opened 2015-10-14
dbs.server.Process.Signal(os.Interrupt) instead of kill , kill causes…

navy1125:
… mongod exit unexpected

dbs.server.Process.Signal(os.Interrupt) instead of kill , kill causes mongod exit unexpected


niemeyer:
Sounds reasonable, thanks.

PR #184, opened 2015-10-29
Bug when a Mongo field name contains a slash

alexcesaro:
I am trying to map a field containing a slash to a struct field using a tag:

```
type user struct {
    FooBar `bson:"foo/bar"`
}
```

But when I do this I get a panic: `Unsupported flag "a" in tag "foo/bar" of type struct ...`

It is caused by a block of code which has this comment: `// XXX Drop this after a few releases.`

Maybe it is time to drop that code?

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -618,24 +618,6 @@ func getStructInfo(st reflect.Type) (*structInfo, error) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


niemeyer:
Yeah, sounds reasonable. Thanks!

PR #185, opened 2015-11-04
Copy the bson.Binary data from the source buffer

gabrielrussell:
If one was unmarshaling from a buffer that gets reused, bson.Binary
values would get corrupted.

### DIFF
--- a/bson/decode.go
+++ b/bson/decode.go
@@ -461,6 +461,7 @@ func (d *decoder) readElemTo(out reflect.Value, kind byte) (good bool) {
 			case typeRawDocElem:
 				out.Set(d.readRawDocElems(outt))
 			}
+			d.readDocTo(blackHole)
 			return true
 		}
 		d.readDocTo(blackHole)


niemeyer:
I'm not entirely sure we want to change the semantics here. Consider the fact we need to allocate an extra buffer for each single binary field we find, instead of using a single byte slice which in the case of the mgo driver was already allocated and will be thrown off if unused.

Also, if we want to make guarantees about not using the source buffer, there's more to be done. Consider Raw, for example. Will we pay the cost of double-allocation for every Raw unmarshalled?

gabrielrussell:
I agree that Raw should contain slices of the input buffer. It clearly makes the most sense. If you were to copy the data when creating a Raw you might recurse and then copy the data again, and so on, completely defeating the usefulness of the Raw.

None of the other simple types wrap slices of the input buffer, and many of them are ref-counted objects that require as much memory as the bindata buffer. This change makes the binary data handled more similarly to how the string data is handled.

If you read in some BSON documents, with binary data fields, into individually allocated buffers, and then marshal them to bson.M variables, then each of the buffers will not be garbage collected until the bson.M variables can be. You're binding the lifecycle of the BSON document buffer to the lifecycle of the unmarshaled data structure. With this change the buffer can be collected as soon as its last regular reference is. If one is creating single use I/O buffers, then my change will reduce memory use, and if one is reusing a single I/O buffer, then my change will make it possible.

Without this change the only reason you can't reuse an I/O buffer to unmarshal in to a bson.M or bson.D is because bson.Binary's reference the I/O buffer. With this change you can reuse an I/O buffer while unmarshaling in any non Raw data structure.

niemeyer:
>  This change makes the binary data handled more similarly to how the string data is handled.

Strings are very inefficient to handle in general because they are immutable, so making it more similar to them isn't a great reason to change it.

> If you read in some BSON documents, with binary data fields, into individually allocated buffers, and then marshal them to bson.M variables, then each of the buffers will not be garbage collected until the bson.M variables can be. You're binding the lifecycle of the BSON document buffer to the lifecycle of the unmarshaled data structure. With this change the buffer can be collected as soon as its last regular reference is. If one is creating single use I/O buffers, then my change will reduce memory use, and if one is reusing a single I/O buffer, then my change will make it possible.

Indeed. The counter side is that your change will make every single binary field become an extra allocation and an extra copy of the data, which means decoding a GridFS file and its 256k chunks will allocate twice as much memory, create twice as much garbage, and become a copying algorithm rather than a zero-copy one.

Without proper analysis I'm not convinced this is a win at all.

PR #196, opened 2015-12-15
correctly skip docs incorrectly being read in to slices

gabrielrussell:
When trying to parse a bson document as a slice, and the slice is not of DocElem or RawDocElem, readElemTo would decide to do nothing, but didn't consume the element, leaving the decoder in a bad state.


niemeyer:
Hey Gabriel,

This doesn't look quite right, and running tests confirms it is indeed introducing a bug. The new logic should be in default, shouldn't it?  Have you tried to run tests there?

Also, we need a test demonstrating why the current logic is broken, and to prevent regressions.

Thanks for the fix.

gabrielrussell:
Yes, absolutely, in the default. I've also added a test.

niemeyer:
Thanks Gabriel.

PR #198, opened 2015-12-18
Apply io.Closer interface for *mgo.Session

osamingo:
I often use `io.Closer` . e.g.

``` go
func AddSignalListenCloser(tag string, closer io.Closer, trigger ...os.Signal) error {

    if tag == "" {
        return fmt.Errorf("singnal: tag should not be empty")
    }

    if closer == nil {
        return fmt.Errorf("singnal: closer should not be nil")
    }

    sigChan := make(chan os.Signal, 1)
    signal.Notify(sigChan, trigger...)

    go wait(sigChan, tag, closer)

    return nil
}

func wait(sigChan chan os.Signal, tag string, closer io.Closer) {

    for {
        <-sigChan
        var err error
        if closer != nil {
            err = closer.Close()
        }
        panic(fmt.Errorf("singnal: %s is close, err = %s", tag, err))
    }

}
```

I want to apply *mgo.Session.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1539,7 +1539,7 @@ func (s *Session) Clone() *Session {
 
 // Close terminates the session.  It's a runtime error to use a session
 // after it has been closed.
-func (s *Session) Close() {
+func (s *Session) Close() error {
 	s.m.Lock()
 	if s.cluster_ != nil {
 		debugf("Closing session %p", s)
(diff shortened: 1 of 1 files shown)


niemeyer:
I appreciate the contribution, but unfortunately that's not something we should be doing. When we return an error, we indicate to developers that this operation can fail, and that the error should be checked. Neither is the case here.

I suggest doing something like [ioutil.NopCloser](https://golang.org/pkg/io/ioutil/#NopCloser) in the affected logic.

Issue #158, opened 2015-08-25
No access to bson.MongoTimestamp components


josler:
Currently there's no easy way to extract the parts of a MongoTimestamp; namely the unix timestamp and the ordinal : http://docs.mongodb.org/master/reference/bson-types/#timestamps

One way to implement this would be to simply add a `Unix() int64` method to MongoTimestamp, which would do some manipulation of the timestamp (int64 -> []byte -> take relevant part as int32 -> int64). This preserves backward compatibility, and whilst there's a few copies needed to get this to work it would be an explicit request.

Another option would be to change the MongoTimestamp type itself and do the parsing on decoding.

Thoughts? - I don't mind submitting a PR for this.

niemeyer:
The timestamp is an internal MongoDB type. May I ask why do you want to unpack it?

If we add it, I'd prefer an interface such as:

    Time() time.Time
    Counter() int32

This matches the current interface on bson.ObjectId.

That said, it'd be nice to have a better picture of why we need this in first.

josler:
Looking to tail the replica set oplog http://docs.mongodb.org/manual/core/replica-set-oplog/; which has a MongoTimestamp field - though I believe the oplog format is an internal implementation detail.

mwmahlberg:
Although it is an internal implementation detail, I see quite some use cases for unpacking the timestamp data. Given the existing backup solutions, it might come in handy to be able to deal with the oplog as it was intended. Furthermore as per the guarantee of uniqueness for a Timestamp on a top level field, I see use cases for message queues and alike. Furthermore, for event storage, unpacking comes in handy for such things as numbers of events within a given second.

mwmahlberg:
Added pull request #202 containing only the necessary commits. Sorry, again.

niemeyer:
There are a few comments there in the pull request. Can you please have a look?

PR #201, opened 2015-12-30
Mongotimestamp parts

mwmahlberg:
This should solve #158 

It provides both `.Time()(time.Time)` and `.Counter()(uint32)`

Additionally, I added a simple generator `NewMongoTimestamp(t time.Time, c uint32)` – although one needs to track the uniqueness manually. I added a crude testable example on how to do that.

If wanted, I could add a more advanced generator.


mwmahlberg:
Sorry, made a mistake including the other commits.

PR #202, opened 2015-12-30
Closes #158 by providing .Time() and .Count()

### LINKED ISSUE #158: No access to bson.MongoTimestamp components

mwmahlberg:
Furthermore, a simple generator for MongoTimestamps is added

Signed-off-by: Markus W. Mahlberg markus.mahlberg@me.com

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -354,16 +354,12 @@ type MongoTimestamp int64
 
 // Time returns the time part of ts which is stored with second precision.
 func (ts MongoTimestamp) Time() time.Time {
-	b := make([]byte, 8, 8)
-	binary.BigEndian.PutUint64(b, uint64(ts))
-	return time.Unix(int64(binary.BigEndian.Uint32(b[:4])), 0)
+	return time.Unix(int64(uint64(ts) >> 32), 0)
 }
 
 // Counter returns the counter part of ts.
 func (ts MongoTimestamp) Counter() uint32 {
-	b := make([]byte, 8, 8)
-	binary.BigEndian.PutUint64(b, uint64(ts))
-	return binary.BigEndian.Uint32(b[4:])
+	return uint32(ts)
 }
 
 // NewMongoTimestamp creates a timestamp using the given date t with second precision
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -41,6 +41,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


bson/bson.go
```diff
@@ -352,6 +352,50 @@ func Now() time.Time {
 // strange reason has its own datatype defined in BSON.
 type MongoTimestamp int64
 
+// Time returns the time part of ts which is stored with second precision.
+func (ts MongoTimestamp) Time() time.Time {
+	b := make([]byte, 8, 8)
+	binary.BigEndian.PutUint64(b, uint64(ts))
+	return time.Unix(int64(binary.BigEndian.Uint32(b[:4])), 0)
```
niemeyer: Isn't that the same as int64(uint64(ts) >> 32)?

bson/bson.go
```diff
@@ -352,6 +352,50 @@ func Now() time.Time {
...
+}
+
+// Counter returns the counter part of ts.
+func (ts MongoTimestamp) Counter() uint32 {
+	b := make([]byte, 8, 8)
+	binary.BigEndian.PutUint64(b, uint64(ts))
+	return binary.BigEndian.Uint32(b[4:])
```
niemeyer: Isn't that the same as uint32(ts)?

niemeyer:
Thanks for the contribution. A couple of comments above.

bson/bson.go
mwmahlberg: Of course. Doh!

mwmahlberg:
Thank you – and of course you are right. Bitshifting is kind of new to me, coming from Java Web Applications.

I'll update

zph:
@mwmahlberg This PR solved an issue I had with needing to create an oplog style `bson.MongoTimestamp`, thank you. The use case I have is generating a bson.MongoTimestamp from time.Time in order to start replaying oplog from point in time calculated using `time.Duration`.

@niemeyer If this is updated to merge cleanly w/ current master, would it be accepted and merged? If so, I'm happy to do the git work/cleanup.

bson/bson.go
```diff
@@ -352,6 +352,46 @@ func Now() time.Time {
...
+	} 
+	
+	tv = uint32(u)
+
+	buf := bytes.Buffer{}
+
+	binary.Write(&buf, binary.BigEndian, tv)
+	binary.Write(&buf, binary.BigEndian, c)
+
+	i := int64(binary.BigEndian.Uint64(buf.Bytes()))
```
zph, line 390: @niemeyer Also, should this style be converted to the terse bitshifting syntax?

mwmahlberg:
Unmaintaned, hence closing

PR #208, opened 2016-01-12
fix bson encode addMap issue

RyouZhang:
add support map[interface{}]interface{} encode

### DIFF
--- a/bson/encode.go
+++ b/bson/encode.go
@@ -129,7 +129,11 @@ func (e *encoder) addDoc(v reflect.Value) {
 
 func (e *encoder) addMap(v reflect.Value) {
 	for _, k := range v.MapKeys() {
-		e.addElem(k.String(), v.MapIndex(k), false)
+		if k.Kind() == reflect.Interface && k.Elem().Kind() == reflect.String {
+			e.addElem(k.Elem().String(), v.MapIndex(k), false)
+		} else {
+			e.addElem(k.String(), v.MapIndex(k), false)
+		}
 	}
 }
 


niemeyer:
This is missing a better description of the problem, and some tests, but perhaps there's a more fundamental problem with the proposal. BSON indeed only supports maps with string keys.. the way we represent that in Go is by enforcing the use of maps with key elements typed as strings, which are also more convenient to use in general. I'm not sure there's much benefit in loosening these rules.

I'm closing down the issue on that basis, but please feel free to complement the report.

RyouZhang:
I understand your concern. But more time use mongodb like this:
{"name":"foo","items":[{"count":1,"price":1.00},{"count":2,"price":2.00}], "total":5.00}
if use map[string]interface{} to convert bson object,
first  golang will judge the slice like []interface{},
second judge interface{} like map[interface{}]interface{},  will trigger the issue

niemeyer:
Sorry, but I still don't understand the issue. Who will introduce the map[interface{}] type?  Do you have sample code?

RyouZhang:
please, try to write this json object to mongodb
{
    "device_model": "iPhone6,2",
    "origin_time": 1453042991466.9,
    "app_build": "20151231",
    "channel": "YOULIN",
    "uid": "8cab9179926546f8a208d5343f851498",
    "local_time": 1453131230831.6,
    "server_time": 1453131100000,
    "country": "中国",
    "city": "杭州",
    "carrier": "中国移动",
    "_id": "10b721fc8dd42b9cd3afcd4434edc485",
    "app_id": "123456",
    "os_ver": "7.1.2",
    "device_id": "8819d7e3cbed8dd6a22ff9629d2f6aa808d1bf2a",
    "os": "iOS",
    "args": {
      "duration": 23950.764894485,
      "last_page_nick": "feed/detail_70db371e040e4647b0f8e166a19bacf7",
      "page_name": "feed/home",
      "page_nick": "feed/home",
      "last_page_name": "feed/detail"
    },
    "community_id": "f2c3572bdca74347b8f1dc00461334d4",
    "imei": "827c939fdca42c6f2550a4843e5fe43c",
    "network": "wifi",
    "app_ver": "2.3.0",
    "screen": "1136x640",
    "event_id": "page_event"
  }

if you define the struct, it works well. but if use map[string]interface{}, it's wrong

niemeyer:
Actual code, please?

RyouZhang:
func (m *MongoInsert) run() {
    var (
        session *mgo.Session
        mh      codec.MsgpackHandle
        item    map[string]interface{}
        name    interface{}
        ok      bool
        err     error
    )
    mh.RawToString = true

```
m.Wg.Add(1)
for raw := range m.Input {
    item = nil
    err = codec.NewDecoderBytes(raw, &mh).Decode(&item)
    if err != nil {
        logrus.WithField("raw", raw).Error(err)
        continue
    }
    name = nil
    for _, key := range m.keys {
        name, ok = item[key]
        if ok {
            name = item[key]
            break
        }
    }
    if name == nil {
        logrus.WithFields(item).Error("invalid collection name")
        continue
    }

    session, err = mongopool.GetMongodbPoolInstance().Get(m.url)
    if err == nil {
        db := session.DB(m.database)
        col := db.C(name.(string))
        err = col.Insert(item)
        if err != nil && false == mgo.IsDup(err) {
            logrus.Error(err.Error())
        }
    } else {
        logrus.Error(err)
    }
}
m.Wg.Done()
```

}
the Actual code

niemeyer:
Sorry for bothering, but can you please provide something I can actually run locally? It should be self-contained and runnable, without using anything that is not relative to the offending package (mgo, bson) or the standard library. Without that there's no way I can tell what the code is in fact doing, or what the types involved are.

RyouZhang:
package main

import (
    "github.com/Sirupsen/logrus"
    "github.com/ugorji/go/codec"
  "gopkg.in/mgo.v2"
)

type JsonByteHook struct {
}

func (j *JsonByteHook) ConvertExt(v interface{}) interface{} {
    return string(v.([]byte))
}
func (j *JsonByteHook) UpdateExt(dest interface{}, v interface{}) {
}

func main() {
    var (
        session *mgo.Session
        jh      codec.JsonHandle
        raw     []byte
        keys    []string
        item    map[string]interface{}
        name    interface{}
        ok      bool
        err     error
    )
    keys = []string{"event", "event_id"}
  raw = []byte("{\"device_model\":\"iPhone6,2\",\"origin_time\":1453042991466.9,\"app_build\":\"20151231\",\"channel\":\"YOULIN\",\"uid\":\"8cab9179926546f8a208d5343f851498\",\"local_time\":1453131230831.6,\"server_time\":1453131100000,\"country\":\"中国\",\"city\":\"杭州\",\"carrier\":\"中国移动\",\"_id\":\"10b721fc8dd42b9cd3afcd4434edc485\",\"app_id\":\"123456\",\"os_ver\":\"7.1.2\",\"device_id\":\"8819d7e3cbed8dd6a22ff9629d2f6aa808d1bf2a\",\"os\":\"iOS\",\"args\":{\"duration\":23950.764894485,\"last_page_nick\":\"feed/detail_70db371e040e4647b0f8e166a19bacf7\",\"page_name\":\"feed/home\",\"page_nick\":\"feed/home\",\"last_page_name\":\"feed/detail\"},\"community_id\":\"f2c3572bdca74347b8f1dc00461334d4\",\"imei\":\"827c939fdca42c6f2550a4843e5fe43c\",\"network\":\"wifi\",\"app_ver\":\"2.3.0\",\"screen\":\"1136x640\",\"event_id\":\"page_event\"}")

```
jh.RawBytesExt = &JsonByteHook{}

err = codec.NewDecoderBytes(raw, &jh).Decode(&item)
if err != nil {
    logrus.WithField("raw", raw).Error(err)
    return
}

for _, key := range keys {
    name, ok = item[key]
    if ok {
        name = item[key]
        break
    }
}

if name == nil {
    logrus.WithFields(item).Error("invalid collection name")
    return
}

session, err = mgo.Dail("mongodb://your mongodb url")
defer session.Close()
if err == nil {
    db := session.DB("your db")
    col := db.C(name.(string))
    err = col.Insert(item)
    if err != nil && false == mgo.IsDup(err) {
        logrus.Error(err.Error())
    }
} else {
    logrus.Error(err)
}
```

}

niemeyer:
As I mentioned, it should be self-contained and runnable, without using anything that is not relative to the offending package (mgo, bson) or the standard library. Without that there's no way I can tell what the code is in fact doing, or what the types involved are.

niemeyer:
Hint: you are using third-party code that does decoding. I bet this is actually what introduces the map[interface{}]interface{} type. If that's the case, then there's nothing wrong with the fact the bson package is strict about accepting map[string] only. That's indeed what's supported by BSON (the spec).

RyouZhang:
ha, it's another interest issue, you can try "encoding/json"，it throw exception can‘t encode map[interface{}]interface{}. the "golang fastjson" do the same, but codec json can encode it. 
the golang judge map[string]interface{}  to map[interface{}]interface{}

niemeyer:
Okay, so mgo/bson is doing exactly the same as encoding/json by telling you that in fact encoded maps must use string keys. I think both packages are correct. If you want to map interface{} keys into string keys, that's easy to do with conversion code.

RyouZhang:
sorry， I don't understand how to do the  conversion code?? Now, when I insert data to mongodb, the mgo/bson throw nothing, but insert error data. 
You can run my code, and check the data in mongodb collection.

Issue #209, opened 2016-01-14
Running bulk after refreshing the session causes panic


szank:
Hi, I just stumbled upon this bug. You can reproduce it more or less like this : 
```
session, _ := 	mgo.DialWithInfo("address")
collection := session.DB("theDB").C("theCollection)
bulk := collection.Bulk()
bulk.Unordered() // I don't think it matters 
[insert data into bulk]

_, err := bulk.Run()
if err != nil{
     //MongoDB is down
     time.Sleep(time.Second)
     session.Refresh()
     bulk.Run() // <- panic 
}
```

stack trace :
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x18 pc=0xb7e6a]

goroutine 1 [running]:
gopkg.in/mgo%2ev2.(*Bulk).runRemove(0xc826c5ae70, 0xc82012f200, 0xc82012f420, 0xc82012f440, 0x13944c)
	/Users/mgalkowski/work/go/src/gopkg.in/mgo.v2/bulk.go:280 +0xca
gopkg.in/mgo%2ev2.(*Bulk).Run(0xc826c5ae70, 0x58, 0x0, 0x0)
	/Users/mgalkowski/work/go/src/gopkg.in/mgo.v2/bulk.go:245 +0x22c
main.runBulkUntilSuccess(0xc820068d00, 0xc826c5ae70)
[snip...]
```

niemeyer:
Fix merged, thanks!

PR #210, opened 2016-01-15
Fix a crash when bulk update and delete returned  an connection error

szank:
Hi, this is a fix for a problem I described in  #209


niemeyer:
Thanks, merging it. I'll add a test for it as well.

PR #212, opened 2016-01-20
Make bson.ObjectId satisfy both encoding.TextMarshaler and encoding.TextUnmarshaler

### LINKED ISSUE #78: BSON ID Values And XML/YAML Marshal/Unmarshal

jackspirou:
This pull request fixes #78 and continues where #79 left off.

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -201,7 +201,6 @@ func readRandomUint32() uint32 {
 	return uint32((uint32(b[0]) << 0) | (uint32(b[1]) << 8) | (uint32(b[2]) << 16) | (uint32(b[3]) << 24))
 }
 
-
 // machineId stores machine id generated once and used in subsequent calls
 // to NewObjectId function.
 var machineId = readMachineId()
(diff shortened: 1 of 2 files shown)


jackspirou:
@niemeyer I just noticed the `bson_test` package is using the package `gopkg.in/mgo.v2-unstable/bson` which seems to be a different repo than `bson`.  Is this current PR good or do the changes made in the `bson` package also need to be added to `gopkg.in/mgo.v2-unstable/bson`?

niemeyer:
Thank you! The package path is fine. gopkg.in/mgo.v2 vs. mgo.v2-unstable are different branches of the same repository. You're targeting the unstable one, so the package path should be unstable as well.

Issue #221, opened 2016-02-08
Querying object that mimics an error value


zippoxer:
I'm using the latest mgo at `gopkg.in/mgo.v2` and MongoDB 3.2.1

When I'm doing this operation (just a basic Apply):

```
var p *SomeStruct
change = mgo.Change{
	Update: bson.M{
		"$set":         p,
		"$setOnInsert": bson.M{"added": time.Now()},
	},
	Upsert:    true,
	ReturnNew: true,
}
info, err := db.C("somecollection").Find(someSelector).Apply(change, p)
```

I get this blank error:
```
&mgo.LastError{Err:"", Code:0, N:1, Waited:0, FSyncFiles:0, WTimeout:false, UpdatedExisting:true, UpsertedId:interface {}(nil), modified:0, ecases:[]mgo.BulkErrorCase(nil)}
```

No message and error code 0, so I can't figure out what's really wrong.

The exact same operation (findAndModify) succeeds through the mongo console.

Why would it return an error when there is none?

Update:

I think figured out the problem.
This is a bug in checkQueryError at session.go. checkQueryError is peeking at the response's raw bytes checking if it has an "err" or "errmsg" before parsing it. Clever trick to avoid parsing, I like it :)
However, my document has a field named "errmsg" and, I guess, my document is included in the response which checkQueryError is searching for "errmsg". When checkQueryError finds "errmsg", it parses the error and since LastError isn't nil (I guess it's unique to successful Apply operations) it returns it as an error, even though there wasn't really an error.

I think the fix would be to check that the LastError's code isn't zero before returning it.

niemeyer:
The test that checks if errmsg exists in the document is just some heuristics to bail early and fast, as that's not for a all query operations. That said, assuming your fix works, that means you have both an "errmsg" value and a "lastErrorObject" value, which means you are storing error values as is, and making it harder for the driver to disambiguate. The fix you propose is probably okay, but it'll have to wait until the next release since this isn't really a recent breakage introduced by the recent driver release. I'd recommend not storing documents that look exactly like an error response if that's possible. Instead, nest them inside a field and everything will be fine. I realize that's suboptimal from an error handling perspective, but the server isn't helping in this case. I hope we can have a much better reporting pipeline in the near future.

zippoxer:
@niemeyer 

I appreciate you paying attention to this.

I'm not storing lastErrorObject. I never knew how lastErrorObject looks like before I dug in session.go to figure out what's wrong.

[...]

PR #222, opened 2016-02-08
fixed LastError check in checkQueryError

zippoxer:
This fixes issue #221.


niemeyer:
Please see conversation in the issue itself.

PR #226, opened 2016-02-23
make decoder can secode string to int

yqylovy:
Now, when the bson type is `string` but the go type is `int`, it will do nothing. I think it's much more convenient and helpful to parse string to int and set to the field.


niemeyer:
This is a bit of a leap, but perhaps it is indeed the most useful thing to do.

Can you please add some tests covering that?

PR #231, opened 2016-02-27
Support Query.SetMaxTime too for count operations

wpjunior:
Currently `SetMaxTime` is not supported for Count operations, this pull request will be inject maxTimeMS if available in count message.


wpjunior:
@niemeyer: any questions about my pull request ?

niemeyer:
No, that looks fine. I'll get it in early this week.

Thanks for the fix!

wpjunior:
@niemeyer, Thanks!

PR #232, opened 2016-02-28
Fix panic on dial error when mongoServer.dial is not set

cezarsa:
I caught this panic after my local MongoDB stopped working during tests and the type assertion failed because conn was nil.


cezarsa:
Oops, I'm sorry. I just found out that this has already been fixed on v2 branch but not merged back to v2-unstable yet.

PR #238, opened 2016-03-04
Fixed error iteration in writeOp

aleist:
The for-range loop incorrectly iterated over lerr.ecases when indexing into oplerr.ecases, causing out of bounds panics if there were errors in previous iterations of the outer loop (which limits the batch size).


niemeyer:
Thank you!

PR #243, opened 2016-03-18
fix the addition index ExpireAfter, for mongo 2.6.10

asm-jaime:
I could not add any ExpireAfter index(no errors), only when fix the division, it worked.
also: https://groups.google.com/forum/#!topic/mgo-users/WV7aparsI1o

### DIFF
--- a/session.go
+++ b/session.go
@@ -1235,7 +1235,7 @@ func (c *Collection) EnsureIndex(index Index) error {
 		Min:              index.Minf,
 		Max:              index.Maxf,
 		BucketSize:       index.BucketSize,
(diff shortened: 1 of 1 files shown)


niemeyer:
That's not quite right, as you can observe by the broken tests.

Please open an issue with a small self-contained reproducer exposig the problem clearly and I'll be happy to help.

asm-jaime:
@niemeyer maybe i'ts my mistake. So, 

> We don't support that because Go ... Note you can set the value to 1 and have the same behavior.

That's mean, a first number what will be correct is 1000000000 time.Duration, or 1 \* time.Second. Any lesser will be equal 0. Maybe best decision: notify about incorrect value.

PR #252, opened 2016-04-12
PurgeMissing now removes invalid transaction tokens

mjs:
In at least one (unhappy) real-world situation, nil values have been been observed in the txn-queue fields of documents. These were causing mgo/txn and PurgeMissing itself to panic.

PurgeMissing now detects and removes all obviously invalid txn tokens (including nils).

### DIFF
--- a/txn/txn.go
+++ b/txn/txn.go
@@ -10,6 +10,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -382,8 +383,8 @@ func (r *Runner) ChangeLog(logc *mgo.Collection) {
 func (r *Runner) PurgeMissing(collections ...string) error {
 
 	type TDoc struct {
-		Id       interface{} "_id"
-		TxnQueue []string    "txn-queue"
+		Id       interface{}   "_id"
+		TxnQueue []interface{} "txn-queue"
 	}
 
 	found := make(map[bson.ObjectId]bool)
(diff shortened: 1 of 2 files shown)


txn/txn.go
```diff
@@ -380,15 +381,23 @@ func (r *Runner) ChangeLog(logc *mgo.Collection) {
...
 // a system that has seen unavoidable corruption back in a working state.
 func (r *Runner) PurgeMissing(collections ...string) error {
-	type M map[string]interface{}
-	type S []interface{}
 
 	type TDoc struct {
-		Id       interface{} "_id"
-		TxnQueue []string    "txn-queue"
+		Id       interface{}   "_id"
+		TxnQueue []interface{} "txn-queue"
```
niemeyer, line 387: That looks a bit awkward. Feels a lot like the data was manually modified to be incorrect, which is something we cannot guard against. Do you have more details of the case?

mjs, line 387: It was seen in a production Juju deployment. AFAIK the customer nor the support team didn't manually make changes to the DB (but I can't be sure of course). There were nil entries in the txn-queue field as well as references to txns which didn't exist in the txns collection (as we've seen elsewhere), indicating data corruption.

This is the only case I know of where we've seen nils in the txn-queue field. I don't have details on what happened to the database or the environment it runs in which might have caused this.

If you're uncomfortable with the change to PurgeMissing we could shelve it for now and dig it out if the problem happens again elsewhere.

mjs, line 387: For the record, I've now personally seen `nil` values in `txn-queue` fields at 2 customer sites now.

PR #267, opened 2016-05-11
Add a Iter.Done function for determining if an Iter is exhausted

ebroder:
We would like to know if an Iter has been exhausted, or if it will have additional results. An Iter may have additional results either if there are more results cached locally, or if we have a non-0 cursor ID. (When mongod returns the last batch of results, it returns a cursor ID of 0).

Conceivably, you could determine this by calling Next and just storing the result until you actually need it, but that's annoying. This is simpler.

The semantics for tail cursors are a bit more complicated (since conceivably you could call HaveMore() and get true, but then call Next() and get false), but tail cursors were always complicated.

I'm not wedded to this specific API, but I would like some way to "peek" at whether there are more results coming down the line. Thoughts?

### DIFF
--- a/session.go
+++ b/session.go
@@ -2332,7 +2332,7 @@ type queryError struct {
 	ErrMsg        string
 	Assertion     string
 	Code          int
-	AssertionCode int        "assertionCode"
+	AssertionCode int "assertionCode"
 }
 
 type QueryError struct {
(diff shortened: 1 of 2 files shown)


niemeyer:
Before anything else, thanks for reaching out.

Can you provide some details about the use case?  It feels like the proper way to handle this would be to simply call Next, as you indicate. If we implement HasNext, we need it to be true. In other words, if HasNext returns true, Next should never return false. This conflicts with the proposed implementation.

ebroder:
My use case is fairly similar to those outlined #170. I have an internal API that's backed by a Mongo query. If the call doesn't return the full result set, I want to set a flag ("there are more results if you ask for them") on the response, hold onto the `Iter`, and giving the client an identifier to reference it.

It basically amounts to a reimplementation of cursors, but for applications where I don't want them to directly connect to (or know about) the Mongo cluster in question. I'd prefer a method like this over calling `Next` - if I do that, I have to store the results and the state tracking generally becomes more complicated. This seemed like a fairly straightforward API change that would make the implementation of my code much simpler.

It looks like several of the other drivers have an "alive" or "dead" method (c.f. [Python](http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.alive) and [PHP](http://php.net/manual/en/mongocursor.dead.php)) which has similar semantics to my implementation - maybe `Alive()` would be a better name for the function?

niemeyer:
Fair enough.

Can we please invert the logic and call the method `Done`. This reflects better the fact we have certainty about it being done, but we have no certainty about really having more data or not (tailable or not).

Then, can we please have one or more tests specific for that functionality that verify that each of these branches is actually doing the right thing, rather than spreading the logic out through other unrelated tests?

Thanks again.

ebroder:
Awesome, that makes sense to me. And sorry for not saying it earlier, but thanks so much for taking a look so quickly!

I've pushed an update that makes the changes you requested to the Iter code. I'll work on adding the tests tomorrow (the initial implementation was mostly to convince me the code worked at all).

One thing I'm not sure about is how to test the `docsToReceive` condition - in order for that test to be reliable, I effectively need to prevent the result of an in-flight operation from getting processed. Do you have any pointers on how to test that particular condition?

niemeyer:
Thank you!

niemeyer:
There's a bug here. iter.err may be set while there's still data available for consumption. I'll need to tweak the implementation.

PR #270, opened 2016-05-18
bson: add AsUpdate

rogpeppe:
Add an AsUpdate function to determine what fields will be
updated by a given document.

We need to to add a document into mongodb, atomically incrementing a
version counter if the document already exists.

The obvious answer is to use Upsert, but to do what we want there, we'd
need to explicitly mention every field in the document (it has quite a
few fields and we add more relatively often).

Say our document type looked like this:

```
    type mdoc struct {
            Id string       `bson:"_id"`
            Field1 int
            Field2 string
            Field3 time.Time       `bson:",omitempty"`
            Field4 string       `bson:",omitempty"`
    }
```

then our code for upserting the document would need to look
something like this:

```
    func addDoc(id string, doc mdoc) {
            set := bson.M{
                    "field1": doc.Field1,
                    "field2": doc.Field2,
            }
            unset := make(bson.M)
            if doc.Field3.IsZero() {
                    unset["field3"] = nil
            } else {
                    set["field3"] = doc.Field3
            }
            if doc.Field4 == "" {
                    unset["field4"] = nil
            } else {
                    set["field4"] = doc.Field4
            }
            c.UpsertId(id, bson.M{
                    "$set": set,
                    "$unset": unset,
                    "$inc", bson.M{"version", 1},
            })
    }
```

We'd much prefer to have the fields automatically derived from
the document type, so when we add a new field, we don't have
to fix the logic and there's less chance of the fields getting
out of step with the type.

There's no great way to do that AFAICS; hence our suggestion
for AsUpdate. With that, we could write the logic as:

```
    func addDoc(id string, doc mdoc) {
            u, err := bson.AsUpdate(doc)
            if err != nil {...}
            delete(u.Set, "version")
            c.UpsertId(id, bson.M{
                    "$set": u.Set,
                    "$unset": u.Unset,
                    "$inc", bson.M{"version", 1},
            })
    }
```

We think that it's a good fit for the bson package itself because
the code to implement AsUpdate depends on the specific
marshaling logic within the bson package, which would
need to be copied if we were to implement the functionality
independently.

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -1580,6 +1580,143 @@ func (s *S) TestObjectIdJSONMarshaling(c *C) {
 	}
 }
 
+var asUpdateTests = []struct {
+	description string
+	obj         interface{}
+	expect      bson.Update
+	expectError string
+}{{
+	obj: struct{ X int }{},
+	expect: bson.Update{
+		Set:   bson.M{"x": 0},
+		Unset: bson.M{},
+	},
+}, {
+	obj: struct {
+		X int `bson:"y"`
+	}{},
+	expect: bson.Update{
+		Set:   bson.M{"y": 0},
+		Unset: bson.M{},
+	},
+}, {
+	obj: struct {
+		X int `bson:",omitempty"`
+		Y string
+	}{
+		Y: "hello",
+	},
+	expect: bson.Update{
+		Set:   bson.M{"y": "hello"},
+		Unset: bson.M{"x": nil},
+	},
+}, {
+	obj: map[string]interface{}{
+		"A": "b",
+		"C": 213,
+	},
+	expect: bson.Update{
+		Set:   bson.M{"A": "b", "C": 213},
+		Unset: bson.M{},
+	},
+}, {
+	obj: &typeWithGetter{
+		result: &typeWithGetter{
+			result: struct{ A int }{1},
+		},
+	},
+	expect: bson.Update{
+		Set:   bson.M{"a": 1},
+		Unset: bson.M{},
+	},
+}, {
+	obj: &bson.Raw{0x03, []byte(wrapInDoc("\x0Aa\x00\x0Ac\x00\x0Ab\x00\x08d\x00\x01"))},
+	expect: bson.Update{
+		Set: bson.M{
+			"a": bson.Raw{0x0a, []byte{}},
+			"c": bson.Raw{0x0a, []byte{}},
+			"b": bson.Raw{0x0a, []byte{}},
+			"d": bson.Raw{0x08, []byte("\x01")},
+		},
+		Unset: bson.M{},
+	},
+}, {
+	obj:         34,
+	expectError: `cannot marshal: Can't marshal int as a BSON document`,
+}, {
+	obj: &typeWithGetter{
+		err: errors.New("some error"),
+	},
+	expectError: `GetBSON failed: some error`,
+}, {
+	obj: &inlineInt{struct{ A, B int }{1, 2}},
+	expect: bson.Update{
+		Set: bson.M{
+			"a": 1,
+			"b": 2,
+		},
+		Unset: bson.M{},
+	},
+}, {
+	obj: &inlineMap{A: 1, M: map[string]interface{}{"b": 2}},
+	expect: bson.Update{
+		Set: bson.M{
+			"a": 1,
+			"b": 2,
+		},
+		Unset: bson.M{},
+	},
+}, {
+	obj:         &structWithDupKeys{},
+	expectError: `Duplicated key 'name' in struct bson_test.structWithDupKeys`,
+}, {
+	obj:         &inlineMap{A: 1, M: map[string]interface{}{"a": 1}},
+	expectError: `Can't have key "a" in inlined map; conflicts with struct field`,
+}, {
+	obj:         &bson.Raw{0x0a, []byte{}},
+	expectError: `cannot marshal: Attempted to marshal Raw kind 10 as a document`,
+}, {
+	obj:         99,
+	expectError: `cannot marshal: Can't marshal int as a BSON document`,
+}, {
+	obj: struct {
+		Id string `bson:"_id"`
+		A  int
+	}{"hello", 1},
+	expect: bson.Update{
+		Set: bson.M{
+			"a": 1,
+		},
+		Unset: bson.M{},
+	},
+}, {
+	obj: map[string]string{
+		"_id": "hello",
+		"a":   "goodbye",
+	},
+	expect: bson.Update{
+		Set: bson.M{
+			"a": "goodbye",
+		},
+		Unset: bson.M{},
+	},
+}, {
+	obj:         map[int]string{34: "hello"},
+	expectError: `map key not a string`,
+}}
+
+func (s *S) TestAsUpdate(c *C) {
+	for _, test := range asUpdateTests {
+		u, err := bson.AsUpdate(test.obj)
+		if test.expectError != "" {
+			c.Assert(err, ErrorMatches, test.expectError)
+		} else {
+			c.Assert(err, Equals, nil)
+			c.Assert(u, DeepEquals, test.expect)
+		}
+	}
+}
+
 type specTest struct {
 	Description string
 	Documents   []struct {
--- a/bson/update.go
+++ b/bson/update.go
@@ -1,0 +1,155 @@
+package bson
+
+import (
+	"fmt"
+	"reflect"
+
+	"gopkg.in/errgo.v1"
+)
+
+// Update represents a document update operation. When marshaled and
+// provided to an update operation, it will set all the fields in Set
+// and unset all the fields in Unset.
+type Update struct {
+	// Set holds the fields to be set keyed by field name.
+	Set map[string]interface{} `bson:"$set,omitempty"`
+
+	// Unset holds the fields to be unset keyed by field name. Note that
+	// the key values will be ignored.
+	Unset map[string]interface{} `bson:"$unset,omitempty"`
+}
+
+// AsUpdate returns the given object as an Update value holding all the
+// fields of x, which must be acceptable to bson.Marshal, with
+// zero-valued omitempty fields returned in Unset and others returned in
+// Set. On success, the returned Set and Unset fields will always
+// be non-nil, even when they contain no items.
+//
+// Note that the _id field is omitted, as it is not possible to set this
+// in an update operation.
+//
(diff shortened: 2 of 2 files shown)


niemeyer:
I've considered a similar idea before, but I'm still slightly uncomfortable with merging something like this alone. It feels too simplistic to be generally advisable. For example, if I have a map of a thousand items and want to insert a new one, this will simply marshal the the whole 1001 items to update the map. I'd prefer people to think about what they are doing rather than believe the driver to be doing the right thing for them when it's not in such cases.

Also, despite having access to the internals, the suggested implementation seems to be doing its logic in a pretty high-level way, by marshaling structs into BSON and then back into a map. If you're willing to go there, the implementation of `AsUpdate` is somewhat simple: marshal in as bson, unmarshal into a map, and parse the struct looking for omitempty tags to separate what you want in set vs. unset.

Finally, that last assumption is worth consideration by itself. The logic interprets an empty field per omitempty logic as meaning "unset it". Very frequently people use structs with omitempty fields precisely because they _don't want_ the field to be touched. This would be the first time the driver is transforming "omitempty means do not marshal it" into "omitempty means the field should be unset".

For all of those reasons, I would like to implement this logic as `bson.Delta(obj1, obj2)`, so the update can then be much richer and more precise.

rogpeppe:
> Also, despite having access to the internals, the suggested implementation seems to be doing its logic in a pretty high-level way, by marshaling structs into BSON and then back into a map

We're actually only marshaling to BSON and back when the object isn't
a struct (in practice that's only for Raw), which is when we can't fetch its
fields directly.

To be honest, the map and Raw cases are only there so that the semantics
are easier to explain - AsUpdate isn't really useful for those cases.
The struct case is the important one (for us, at any rate) and I'd be
quite happy if the logic was restricted to structs only.

> Very frequently people use structs with omitempty fields precisely because they don't want the field to be touched.

Interestingly, one place where the field _is_ touched is when the document
is overwritten, because only the fields that are included will be in
the resulting document.  That's the case we're interested here - we want
to maintain the invariant that fields with the omitempty qualifier
are never in the database with their zero value.

> For all of those reasons, I would like to implement this logic as bson.Delta(obj1, obj1), so the update can then be much richer and more precise.

Delta sounds attractive but doesn't quite address our requirements, because
I think that in order to get any "unset" operations, the fields in the
first object would need be non-zero, and that means that
one would need to keep some object in sync with the type
containing arbitrary non-zero values for all fields.

That is, bson.Delta(T{}, x) (where x is of type T) could never return
an unset operation AFAICS because all fields in T would be zero.

We've been through a few iterations of function name and semantics.
Other suggestions appreciated.

rogpeppe:
> parse the struct looking for omitempty tags to separate what you want in set vs. unset.

That's some of the logic that's in mgo/bson that we'd prefer to avoid second-guessing
in third party code. It's not totally trivial to do that in the presence of inline and it would
be nice not to duplicate the type cache too, simple though it is.

niemeyer:
> We're actually only marshaling to BSON and back when the object isn't
> a struct (in practice that's only for Raw), which is when we can't fetch its
> fields directly.

I see. That sounds okay.

> Delta sounds attractive but doesn't quite address our requirements, because I think that in order to get any "unset" operations, the fields in the first object would need be non-zero, and that means that one would need to keep some object in sync with the type containing arbitrary non-zero values for all fields.

We might support a mode with Delta(nil, obj1), which is closer to your requirements.

> That is, bson.Delta(T{}, x) (where x is of type T) could never return an unset operation AFAICS because all fields in T would be zero.

Well, of course, because obj1 has every field unset, so there's nothing to unset. Along similar lines, Delta(x, x) will always be empty because the value is the same. These are well understood semantics for delta operations, which is most of the reason why it's a good API. The rest of the reason is that this can much more efficiently represent the delta for streaming.

rogpeppe:
> We might support a mode with Delta(nil, obj1), which is closer to your requirements.

So Delta(nil, obj) would be equivalent to AsUpdate(obj) ?

That would seem reasonable.

Implementing Delta itself efficiently without doing any marshaling/unmarshaling round trippage
seems hard, because deep checking equality is going to be hard.

I can imagine a version of Delta implemented in terms of (an unexported) AsUpdate, something
like this:

```
// Delta returns the updates that will result if the object
// obj0 is updated to be obj1. It does not always return
// the most efficient update - when the values are hard
// to compare, it might result in redundant Set members.
//
// If obj1 is nil, the result will be as if obj0 was a value
// of the same type as obj0 with all fields non-zero and
// different from those in obj1.
func Delta(obj0, obj1 interface{}) (Update, error) {
    u1 := asUpdate(obj1)
    if obj0 == nil {
        return u1, nil
    }
    u0 := asUpdate(obj0)

    for attr := range u1.Unset {
        if _, ok := u0.Unset[attr]; ok {
            delete(u1.Unset, attr)
        }
    }
    for attr, val1 := range u1.Set {
        if val0, ok := u0.Set[attr]; ok && obviouslyEqual(val0, val1) {
            delete(u1.Set, attr)
        }
    }
    return u1
}

// obviouslyEqual returns true if x and y are easily comparable
// and will be equal when marshaled as BSON, and false
// otherwise.
func obviouslyEqual(x, y reflect.Value) bool {
    ...
}
```

I guess the obviouslyEqual function could fall back to marshaling and comparing
bytes for the harder cases.

I'm interested in your envisaged use cases for Delta - I haven't yet come across a situation
where something like that might be useful.

niemeyer:
I really doubt you haven't.

Very frequently we have an object at hand that we want to change a field for in the database:

```
// ... arbitrary changes in obj ...
... := collection.Update(bson.M{"$set": ...obj.field...})
```

The proposed idea allows transforming that logic into:

```
old := obj
// ... arbitrary changes in obj ...
... := collection.Update(bson.Delta(old, obj)))
```

rogpeppe:
I like that idea, but I'm not sure how it can work decently. As a really simple example,  say I have the type:

```
type T struct {
    A int
    // other fields
}
```

and I have a method that wants to set A in the database. 

```
func SetA(a int) {
    old := T{ ??? }
    obj := T {A: a}
    ... := collection.Update(bson.Delta(old, obj)))
}
```

What should we set the value of old to if a can be any int? ISTM that the only way is to set old.A to any value other than a (a+1 for example) because otherwise we are not guaranteed that a will be set (or unset if A is marked as omitempty).

It feels to me like it would be more awkward in practice than just writing out the update doc.

rogpeppe:
Another possibility might be to go with AsUpdate (or other name with similar semantics) but add arguments to explicitly include and/or exclude a set of fields from the result.

niemeyer:
Delta would return a rich update statement transforming the first object into the second. The first object would represent the result of a query in general, rather than being hand cooked. If the first object is nil, it would interpret the intention to be to blast whatever is in place and make it equal to inserting the second object, which is what you want.

rogpeppe:
> The first object would represent the result of a query in general, rather than being hand cooked.

I don't understand how that could work without being racy. First, you probably don't want to do a query before doing an update as it's an unnecessary extra round trip to MongoDB. Second, the value you just retrieved might not be anything like the value that's in the database when you actually apply the update.

But I'm sure I misunderstand you somehow. Perhaps you could show how an existing piece of code could be changed to use Delta?

niemeyer:
It seems somewhat self-evident that updates to database documents frequently consider a prior state that is in memory. And yes, it's racy if there are concurrent changes in progress. Every single update has the potential to be racy, inherently. If that's a problem then the interaction design is broken.

But please don't worry about those details. It's not a problem you have and so there's no need to be concerned with it. We just cannot merge your proposal because it has the issues described above, sorry.

rogpeppe:
> Every single update has the potential to be racy, inherently.

Not _that_ racy.  Consider a document where the fields must be
self-consistent and updated together.

```
type doc struct {
    X, Y int
}
```

We have two concurrent actors that are both concurrently trying to set
different (self-consistent) values for X and Y and they're using Delta
to do it.  The document has the initial value {X: 1, Y: 2}.

Alice: get -> {X: 1, Y: 2}
Bob: get -> {X: 1, Y: 2}

Alice: set delta({X: 1, Y: 2}, {X: 3, Y: 2}) -> $set {x: 3}
Bob: set delt({X: 1, Y: 2}, {X: 1, Y: 4}) -> $set {y: 4}

The final value is {X: 3, Y: 4}, which is not self-consistent
and doesn't represent the value required by either of the
actors.

If both had done $set with values for both x and y, then
although there's a race, the final value will at least be internally
consistent with one of the updates.

That's an inherent problem with Delta as you've suggested it
AFAICS.

niemeyer:
You basically agreed with me in a very wordy way. Update operations have the potential to be racy. Delta simply computes the delta automatically.. it doesn't make it more racy or less racy.

rogpeppe:
> Update operations have the potential to be racy. Delta simply computes the delta automatically.

It depends what you mean by "racy". There's the inherent race involved
in two updates running concurrently - they might execute in either order
and that's fine because the end result is either one or the other. There's
also the race that results from two updaters reading a value, modifying
it and writing it back, which is not fine - we can end up with a value
that's neither of the two intended values. We wouldn't try to increment a value
in the database that way - why should Delta be any different?

niemeyer:
I know you are a very smart person, Roger, easily capable of understanding that a $set that is generated by a delta and a $set that is generated by hand has exactly the same potential for race conditions.

Knowing you personally, I also understand you love to argue about simple points. We don't have to do that.

rogpeppe:
> a $set that is generated by a delta and a $set that is generated by hand has exactly the same potential for race conditions.

No. Assuming no conditional logic, a $set generated by hand will always set the same number of fields. A $set generated by a delta will set some number of fields depending on the value presented as the "before" value of the delta. The former will set all the required fields consistently. The latter may not.

I am honestly trying to understand how I might use Delta in a useful way in real code. A example of how you might change some real existing code to use it would be helpful. You may see this as a simple point, but I don't, sorry.

niemeyer:
Sorry, but I consider the reasoning above enough for a good understanding of the problem.

Thanks for the details you provided and the code you contributed. It may end up being used.

rogpeppe:
FWIW I've discussed with some others, and they can't understand the reasoning either. Sometimes a fragment of code can provide greater understanding than prose.

A parrticular phrase of your explanation that eludes me is "the result of a query in general". If I understood what you meant by that,  perhaps I'd understand your position (which I'd like to, BTW - I'd much prefer to contribute to the code base rather than forking parts of it).

niemeyer:
I mean it. :-)

PR #273, opened 2016-05-22
Godoc comment updates for default values

### LINKED ISSUE #104: documentation should mention default consistency and safe modes
It appears that the documentation does
not mention what default values are used when SetSafe and SetMode
are not called. If I've missed this, the docs should be changed
so a search for "default" will find it (I did not read through the
entire doc in detail). If I haven't missed it, then the docs should
specify the default or say explicitly that the default is undefined.

sha1sum:
The `SetMode`, `SetSafe`, and `SetSocketTimeout` functions' documentation comments have been updated to show their default values.

This resolves #104

### DIFF
--- a/session.go
+++ b/session.go
@@ -1580,7 +1580,8 @@ func (s *Session) Refresh() {
 	s.m.Unlock()
 }
 
-// SetMode changes the consistency mode for the session.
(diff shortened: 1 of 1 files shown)


niemeyer:
Thank you!

PR #274, opened 2016-05-24
Unmarshal returns time.Time in UTC.

alesstimec:
Documentation suggests that timestamps will be in UTC but that doesn't appear to be the case, this PR addresses that issue.

### DIFF
--- a/bson/bson_test.go
+++ b/bson/bson_test.go
@@ -148,7 +148,7 @@ var allItems = []testItemType{
 		"\x08_\x00\x00"},
 	{bson.M{"_": true},
 		"\x08_\x00\x01"},
-	{bson.M{"_": time.Unix(0, 258e6)}, // Note the NS <=> MS conversion.
+	{bson.M{"_": time.Unix(0, 258e6).UTC()}, // Note the NS <=> MS conversion.
 		"\x09_\x00\x02\x01\x00\x00\x00\x00\x00\x00"},
 	{bson.M{"_": nil},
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks for the contribution, but I'm not quite sure this is the right thing to do. Timestamps in the database do not have a timezone, and people often work with timestamps in their own timezones.

Which documentation suggests it is UTC?

alesstimec:
Mongo documentation states all timestamps are stored in UTC (https://docs.mongodb.com/manual/tutorial/model-time-data/). So it is very unexpected when i retrieve objects from mongo and times are in Local. It also makes testing a lot harder since we put in times in UTC, but have to do jc.DeepEquals with timestamps in Local..

niemeyer:
Right, that documentation describes stamps should be stored in UTC, and mgo does store times as UTC per the specification. It also does not corrupt the stored times pretending they are in the local timezone either. Instead, it recognizes the fact the the stored timestamp is UTC, and loads that timestamp under the local timezone per normal Go rules. In practice, that's exactly the same you get when you run `time.Now`, or `time.Unix`: a stamp that is timezone aware, and may be easily set to UTC if you call the respective method.

So it sounds like the best possible outcome with the rules available: applications can use timestamps as known locally per default Go behavior, and these can roundtrip through the database as UTC using both the proper conventions of MongoDB and Go.

It's not clear to me what exactly you're struggling with or trying to fix.

PR #278, opened 2016-06-08
Update godoc, GridFS chunck size default value changed

chapsuk:
[MongoDB doc.](https://docs.mongodb.com/manual/core/gridfs/#files.chunkSize)


niemeyer:
Thanks, I'll fix the doc. It's actually always 255kb, because it's a client-side defined setting.

PR #282, opened 2016-06-17
Added function that Marshals BSON.M to JSON

ajpen:
Its pretty much a wrap of the json.Marshal function, somewhat integrating it into package. I don't know how useful it would be but I prefer having it in the source than having to build the function in the application I have thats using mgo/bson.


ajpen:
Can see this as unnecessary. Closing...

PR #285, opened 2016-06-21
dont mix input and output buffers in cgo calls

gabrielrussell:
Since go 1.6 now disallows go pointers to go pointers passed to cgo functions, the sasl code was broken on windows. The use of a single pointer as an input buffer and output buffer is basically illegal. This separates the buffers in to separate input and output buffers.


niemeyer:
Thanks, Gabriel.

PR #286, opened 2016-06-21
dont mix input and output buffers in cgo calls

gabrielrussell:
This is just a backport from v2-unstable. Since the code in v2 always fails on windows under 1.6, this should be fixed sooner then later. If a merge from v2-unstable is upcoming, then discard this pull request.


mpobrien:
The other PR making the corresponding change to v2-unstable was already merged (https://github.com/go-mgo/mgo/pull/285) which was then merged back to v2, so I think this PR can also just get closed as a no-op.

Issue #277, opened 2016-06-03
duplicate key error when inserting into txns.stash on mongo 3.2


babbageclunk:
With mongo 3.2, running two conflicting transactions (both trying to insert the same document) in parallel sometimes produces a duplicate key error instead of one of the transactions failing with `txn.ErrAborted`.

The error looks like this:
```
&mgo.QueryError{
    Code:11000, 
    Message:"E11000 duplicate key error collection: test.txns.stash index: _id_ dup key: { : { c: \"target\", id: \"doc-id\" } }", 
    Assertion:false,
}
```

I've attached a program that shows this - on my machine it takes about 40-60 times around the loop to see the error. You can run it with `go run mgo-duplicate-key.go <dir containing mongod 3.2 binary>`

The transactions both need to make an assertion about another existing document before trying to insert the conflicting document - without the first assertion the failure doesn't happen.

[mgo-duplicate-key.go.zip](https://github.com/go-mgo/mgo/files/297483/mgo-duplicate-key.go.zip)

PR #291, opened 2016-06-27
Retry upserts that fail with "duplicate key error"

babbageclunk:
According to the Mongo docs (link below), getting this error from an upsert is expected
behaviour and the client should retry if it happens.
https://docs.mongodb.com/v3.2/reference/method/db.collection.update/#use-unique-indexes

It doesn't seem to be specific to Mongo 3.2, but I think that the performance characteristics have changed in a way that makes it much more common than it is with 2.x.

Looking through the code revealed that removes had the same problem - I verified this by modifying the repro code to do removes instead.

This fixes https://github.com/go-mgo/mgo/issues/277


niemeyer:
Any chance we can get a test?

Also, wonder if we should be retrying on mgo itself. Any reason not to?

babbageclunk:
Good point, the retry should definitely be in mgo, rather than mgo.txn. I'll move it. Are you alright with the infinite loop to do the retry? Or should I be doing something a bit cleverer?

Sorry, I should have added a test - it's fiddly to do, but it looks like there are already some other tests trying to reproduce racy things.

babbageclunk:
Hi, sorry for the delay on this - I've moved the retrying into mgo, in Query.Apply and Collection.Upsert. Unfortunately I've had real trouble reproducing the problem in a test at that level - no matter how I try I can't get the upserts to fail with an 11000 error. 

However when I run the txn tests TestTxnQueueStashStressTest was reliably failing and now passes consistently with the change. (I don't really understand why the txn tests aren't run by `go test ./...` from the mgo directory.)

babbageclunk:
It seems like the -check.v in .travis.yml means the mgo/txn tests aren't being run - is that deliberate? I've added a line to run the txn tests as well.

babbageclunk:
Hmm - that's unfortunate. It looks like TestTxnQueueStressTest fails intermittently on the parent branch as well. I'm going to pull the change to turn on the mgo/txn tests out of this PR into a new one and add a skip for that test until someone can work out what's going wrong.

babbageclunk:
Sigh - no, of course, that doesn't work because this PR fixes TestTxnQueueStashStressTest. So I'm including the skip here.

babbageclunk:
To show that this change does fix the duplicate key error, [here's](https://travis-ci.org/babbageclunk/mgo/builds/142854817) a build of a branch that has the txn tests turned on without the upsert retrying - TestTxnQueueStashStressTest fails in the mongo 3.2 test run.

session.go
```diff
@@ -2484,7 +2484,15 @@ func (c *Collection) Upsert(selector interface{}, update interface{}) (info *Cha
 		Flags:      1,
 		Upsert:     true,
 	}
-	lerr, err := c.writeOp(&op, true)
+	var lerr *LastError
+	for {
```
niemeyer: Let's please start by retrying at most 5 times here and in the other loop. More than that there's a good chance that there's something else wrong.

session.go
```diff
@@ -4240,12 +4248,20 @@ func (q *Query) Apply(change Change, result interface{}) (info *ChangeInfo, err
...
+		if err == nil {
+			break
+		} else if change.Upsert && IsDup(err) {
+			// Retry duplicate key errors on upserts.
+			// https://docs.mongodb.com/v3.2/reference/method/db.collection.update/#use-unique-indexes
+			continue
+		} else if qerr, ok := err.(*QueryError); ok && qerr.Message == "No matching object found" {
 			return nil, ErrNotFound
+		} else {
+			return nil, err
```
niemeyer: Please take this one out of the else and after the if blocks. Makes it more clear that this isn't looping around no matter what.

session.go
```diff
@@ -4240,12 +4248,20 @@ func (q *Query) Apply(change Change, result interface{}) (info *ChangeInfo, err
...
+	for {
+		err = session.DB(dbname).Run(&cmd, &doc)
+
+		if err == nil {
+			break
+		} else if change.Upsert && IsDup(err) {
+			// Retry duplicate key errors on upserts.
+			// https://docs.mongodb.com/v3.2/reference/method/db.collection.update/#use-unique-indexes
+			continue
+		} else if qerr, ok := err.(*QueryError); ok && qerr.Message == "No matching object found" {
```
niemeyer: Not your problem, but the error message comparison here saddens me.

txn/flusher.go
```diff
@@ -264,8 +264,7 @@ NextDoc:
 		// Document missing. Use stash collection.
 		change.Upsert = true
 		chaos("")
-		_, err := f.sc.FindId(dkey).Apply(change, &info)
-		if err != nil {
+		if _, err := f.sc.FindId(dkey).Apply(change, &info); err != nil {
```
niemeyer: Please restore the original form here. No need to touch this file.

txn/mgo_test.go
```diff
@@ -17,16 +21,39 @@ type MgoSuite struct {
 	session *mgo.Session
 }
 
-var mgoaddr = "127.0.0.1:50017"
+const mgoip = "127.0.0.1"
```
niemeyer: The changes in this file seem unrelated to the fix. Note that most of the mgo tests use fixed ports for the tests, so we have larger problems if that was an issue.

Also, we have a test server package nowadays which could handle this logic all by itself, probably. So if we have to fix it, that'd be a better way.

session.go
babbageclunk: Done

session.go
babbageclunk: Done

txn/flusher.go
babbageclunk: Oops, that was a holdover from the previous version of the PR. Removed.

txn/mgo_test.go
babbageclunk: You're right - these changes are only tangentially related to the upsert fix. Would you rather I split them (and the travis config change) out into a separate PR? It would need to get merged after this one, otherwise the txn tests will be failing.

 D'oh, I hadn't spotted it, dbtest handles this much more nicely. I'll rewrite txn_test.go to use it and remove mgo_tes.go.

babbageclunk:
Thanks for going through this!

babbageclunk:
This is a spurious failure - the script to start the various mongo processes and set up users failed with `couldn't add user: not master`
The same commit passed on Travis here: https://travis-ci.org/babbageclunk/mgo/builds/143769207

session.go
```diff
@@ -4240,9 +4251,16 @@ func (q *Query) Apply(change Change, result interface{}) (info *ChangeInfo, err
...
 	var doc valueResult
-	err = session.DB(dbname).Run(&cmd, &doc)
-	if err != nil {
-		if qerr, ok := err.(*QueryError); ok && qerr.Message == "No matching object found" {
+	for i := 0; i < maxUpsertRetries; i++ {
+		err = session.DB(dbname).Run(&cmd, &doc)
+
+		if err == nil {
+			break
+		} else if change.Upsert && IsDup(err) {
```
rogpeppe: given that the else is redundant, perhaps leave it out and put the if on a new line?

session.go
rogpeppe: ditto about the else.

txn/txn_test.go
```diff
@@ -689,7 +702,12 @@ func (s *S) TestPurgeMissingPipelineSizeLimit(c *C) {
 	c.Assert(err, IsNil)
 }
 
+var flaky = flag.Bool("flaky", false, "Include flaky tests")
+
 func (s *S) TestTxnQueueStressTest(c *C) {
+	if !*flaky {
```
rogpeppe, line 715: Perhaps include a comment as to why this test is flaky and how it fails when it flakes out?

rogpeppe:
LGTM FWIW. What an unfortunate semantic.

rogpeppe:
BTW it would be great if this could land soon, as it fixes a current critical bug in juju.

session.go
babbageclunk: Removing those two elses and making the ifs independent reads much better, thanks

session.go
```diff
@@ -2484,7 +2487,15 @@ func (c *Collection) Upsert(selector interface{}, update interface{}) (info *Cha
 		Flags:      1,
 		Upsert:     true,
 	}
-	lerr, err := c.writeOp(&op, true)
+	var lerr *LastError
+	for i := 0; i < maxUpsertRetries; i++ {
```
jameinel, line 2491: I believe mgo has an ability to hook in a log function, can we get retries like this into that mechanism so we have an idea how often it might be triggering.
Maybe something like logging only on success if i > 0 ?

babbageclunk, line 2491: Done

session.go
```diff
@@ -2484,7 +2487,18 @@ func (c *Collection) Upsert(selector interface{}, update interface{}) (info *Cha
...
+		lerr, err = c.writeOp(&op, true)
+		// Retry duplicate key errors on upserts.
+		// https://docs.mongodb.com/v3.2/reference/method/db.collection.update/#use-unique-indexes
+		if !IsDup(err) {
+			if i > 0 {
+				debugf("upsert retry succeeded after %d failure(s)", i)
```
niemeyer, line 2497: The error message is bogus. Not having a dup error does not mean the upsert succeeded.

Please drop the debug message altogether. If we have debug on, we'll see the attempts going through.

session.go
```diff
@@ -4240,8 +4254,20 @@ func (q *Query) Apply(change Change, result interface{}) (info *ChangeInfo, err
...
 
 	var doc valueResult
-	err = session.DB(dbname).Run(&cmd, &doc)
-	if err != nil {
+	for i := 0; i < maxUpsertRetries; i++ {
+		err = session.DB(dbname).Run(&cmd, &doc)
+
+		if err == nil {
+			if i > 0 {
+				debugf("upsert retry succeeded after %d failure(s)", i)
```
niemeyer, line 4262: Same, let's please drop this.

niemeyer:
Merging, will fix that afterwards.

babbageclunk:
Thanks!

babbageclunk:
I've got a branch with these same changes against the v2 branch. Do you want me to make another PR for that, or will you just merge this across to there?

PR #294, opened 2016-07-05
Increase bson.NewObjectId() performance by caching the process id

256dpi:
Every call to `bson.NewObjectId()` will also call `syscall.Getpid()` to obtain the process id. This PR caches the id and increases the performance:

```
Before:
PASS: bson_test.go:1828: S.BenchmarkNewObjectId 10000000  228 ns/op

After:
PASS: bson_test.go:1828: S.BenchmarkNewObjectId 20000000 76.0 ns/op
```

AFAIK go does not support process forking, the process id should therefore always be the same.

I noticed this while working on a system that writes a lot of data to the database and generates the ObjectId on the client. This PR significantly increases the performance of the said system.

### DIFF
--- a/bson/bson.go
+++ b/bson/bson.go
@@ -205,6 +205,7 @@ func readRandomUint32() uint32 {
 // machineId stores machine id generated once and used in subsequent calls
 // to NewObjectId function.
 var machineId = readMachineId()
+var processId = os.Getpid()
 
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks!

256dpi:
Thanks! How long will it take until this change is merged into v2 (stable)?

PR #295, opened 2016-07-06
Fix SetCursorTimeout. See https://jira.mongodb.org/browse/SERVER-24899

BenLubar:
`Tailable`, `AwaitData`, and `NoCursorTimeout` were not set, and for some reason MongoDB has two copies of each flag. And it was looking at the other set.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1581,7 +1581,7 @@ func (s *Session) Refresh() {
 }
 
 // SetMode changes the consistency mode for the session.
-// 
+//
 // The default mode is Strong.
 //
 // In the Strong consistency mode reads and writes will always be made to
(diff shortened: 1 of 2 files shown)


BenLubar:
@niemeyer it looks like one of the Travis builds failed because a port was already in use.

niemeyer:
Thanks!

Just restarted the failing test.

Speaking of tests, can we have a test that breaks so we can be sure these changes are doing what we want them to do in the future?

BenLubar:
You had a test case for cursors that are supposed to time out, but no test case for cursors that aren't supposed to time out.

niklaskorz:
Any reason this hasn't been merged yet? The behavior for no timeout can't get any more broken than it currently is, no matter if there's a test for no timeout.

bigodines:
Is there anything missing to this PR to get it merged?

PR #300, opened 2016-07-18
don't crash if password not specified for sasl acquirecredentials

mpobrien:
This will avoid crashing when`password` is null.


niemeyer:
Thanks!

PR #302, opened 2016-07-21
Handle upsert duplicate key errors correctly if retries run out

babbageclunk:
The handling of the duplicate key error was wrong if the maximum number of upsert retries was reached. The final error would be ignored.

Restructure the upsert retry logic to account for this.

Change the interpretation of maxUpsertRetries to exclude the first attempt, which makes the name clearer.

Also remove unwanted logging about retries.

### DIFF
--- a/session.go
+++ b/session.go
@@ -147,7 +147,10 @@ var (
 )
 
 const (
-	defaultPrefetch  = 0.25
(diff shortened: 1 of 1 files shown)


niemeyer:
This was fixed and tested via #316.

Issue #313, opened 2016-08-10
ErrNotFound instead of Duplicate Error on upsert breaks compatibility


OSadovy:
Hi,

Part of our app uses mongo db for distributed lock, which broke with recent build of mgo/v2. I nailed down the problem to merge of pull request #291 .

Example:
```
func mongoLock() (bool, string, error) {
	dt := time.Now().UTC()
	var (
		lock         bson.M
		prevInstance string
	)
	_, err := rm.db.C("systemlocks").Find(
		bson.M{"_id": "all", "dt": bson.M{"$lt": dt.Add(-LockInterval)}},
	).Apply(mgo.Change{
		Update: bson.M{"$set": bson.M{"dt": dt, "instance": InstanceName}},
		Upsert: true,
	}, &lock)
	if err != nil {
		if mgo.IsDup(err) {
			return false, "", nil
		}
		return false, "", err
	}

	if _, ok := lock["instance"]; ok {
		prevInstance = lock["instance"].(string)
	}
	return true, prevInstance, nil
}

func mongoUnlock() error {
	dt := time.Now().UTC()
	err := rm.db.C("systemlocks").Update(
		bson.M{"_id": "all", "instance": InstanceName, "dt": bson.M{"$lte": dt}},
		bson.M{"$set": bson.M{"dt": time.Time{}}},
	)
	if err != nil {
		if err == mgo.ErrNotFound {
			log.Error("Error releasing lock: not acquired!")
			return nil
		}
		return err
	}
	return nil
}
```

Earlier mgo returned duplicate error when another instance held the lock, but now, I assume it retries and returns ErrNotFound. Reading mongo documentation which was referenced in the above pull request, it says "If the operation fails because of a duplicate index key error, applications may retry the operation which will succeed as an update operation."

*May* is not *should*, so maybe you reconsider the decision to retry *every* upsert operation silently?

niemeyer:
That's certainly interesting. Can you provide more details about why you are trying to catch an ErrNotFound on an upsert, given that the typical semantics is "insert if it does not exist", rather than returning "not found"?

That's the key of the problem. It is indeed incompatible with the previous behavior, but my reading is that this was actually a bug fix for pretty much every application relying on mgo today. I've never seen myself people expecting that error on their upserts. Typically it just breaks down.

OSadovy:
@niemeyer,

Actually, as per my code example, I am expecting dup error, not ErrNotFound.
```
if mgo.IsDup(err) {
[...]

PR #315, opened 2016-08-11
Fix the IsDup retry error handling on Apply.

niemeyer:
Reported by Cezar Sá Espinola in #313.


mjs:
Note that #302 also fixes this issue and in an arguably more elegant way.

niemeyer:
This pulls in too many unrelated changes. Will merged later by doing v2 => v2-unstable.

PR #316, opened 2016-08-11
Fix the IsDup retry error handling on Apply.

niemeyer:
Reported by Cezar Sá Espinola in #313.


mjs:
Note that #302 also fixes this issue and in an arguably more elegant way.

PR #328, opened 2016-08-30
Add an iter.NeedsNetwork function

alanhdu:
We have a tailable cursor of a capped collection with lots of inserts. Right now, using a `for tail.Next(&result) {...}` mgo issue roughly one getmore request per insert, which means our mongod instance can get 100s of getmores per second.

We'd like a way to tell mgo to wait a little before issuing a getmore so we can "batch" documents together. Adding a `iter.NeedsNetwork` function, which tells us whether the next `iter.Next` call will issue a getmore request seemed the like easiest way to do this.

I'm not attached to the API or the name, but I'd like some way to cut down on the number of requests mgo issues.

### DIFF
--- a/session.go
+++ b/session.go
@@ -3649,6 +3649,22 @@ func (iter *Iter) Done() bool {
 	}
 }
 
+// NeedsNetwork returns true if only if a follow up Next call will issue
+// a getMore operation to mongo
+func (iter *Iter) NeedsNetwork() bool {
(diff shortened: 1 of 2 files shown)


evan-stripe:
@niemeyer Err, sorry to bug, but any thoughts on this? This would be pretty helpful for one of our use cases.

PR #339, opened 2016-09-27
Added LazyConnection option

mjgarton:
Allow a Session to be returned before we've established a connection to
mongodb.  This allows applications easily to start up even when mongodb
is unavailable and to start working properly when the database becomes
available.


ghost:
After creating this, I noticed that a similar PR already exists: https://github.com/go-mgo/mgo/pull/281

PR #360, opened 2016-11-29
txn: return error from flusher.checkpoint

axw:
If the call to f.tc.Update in flusher.checkpoint
fails due to an error other than mgo.ErrNotFound,
the error is currently swallowed. flusher.advance
will then loop back around and call checkpoint
again, which may continue to fail (e.g. if the
Mongo connection is in a bad state), pegging the
CPU.


rogpeppe:
Please add a test case for this issue and verify that it fails without this fix.

PR #369, opened 2016-12-14
bson: check utf8 validity

bpot:


bson/encode.go
```diff
@@ -294,6 +295,9 @@ func (e *encoder) addElem(name string, v reflect.Value, minSize bool) {
 			}
 		default:
 			e.addElemName(0x02, name)
+			if !utf8.ValidString(s) {
+				panic("string value is not valid utf8")
```
bpot, line 299: panics are captured and returned as errors from `Marshal`

PR #370, opened 2016-12-15
Configurable near ping threshold

Oppodelldog:
During load tests, when firing a lot of queries against a replica-set one of the servers will be detected not to be near (15ms) in BestFit-method.  
This leads to a swap of all queries to the seconds slave which subsequently will become slow as well.  
After the next cluster sync the load will swap to the first slave again and some kind of ping-pong balancing occurs.
To fine-tune the BestFit-Method I implemented an option to overwrite the near-threshold for the session.

I'd like to hear your opinion on this change.


Oppodelldog:
PR close since this project is not maintained anymore, thanks for all the effort

PR #373, opened 2016-12-16
Encoding/Decoding time.Duration to/from int64

smoya:
This PR allows to convert properly a `time.Duration` to milliseconds, following the approach of Date bson type (stores in milliseconds too) in order to be able to add or subtract durations in mongo.


gonzaloserrano:
👍

PR #378, opened 2016-12-30
mongoServer drops abended state after `ismaster` check

g7r:
There is no way for `mongoServer` to leave `abended=true` state. Which is somewhat bad. `mongoCluster` performs `ismaster` check after each `server.AcquireSocket(...)`. `ismaster` check flushes all cached credentials on acquired socket. And Mongo auth is not cheap CPU-wise. Authenticating for each Mongo server request rockets up CPU load.


g7r:
Bye, mgo

Issue #308, opened 2016-08-01
Add support for normalization field in Collation


jyemin:
MongoDB 3.4 collation support includes a "normalization" field of type boolean.  It should be added to the new Collation struct.

niemeyer:
Yeah, as mentioned I expected that given it was in the actual index, despite not being in the design doc.

No problems, though. Will be on the next stable.

jyemin:
Also required is a maxVariable field of type string.

jyemin:
maxVariable: Controls which characters are affected by alternate: “shifted”.

   “punct” (punctuation and spaces are affected)
   “space” (only spaces)

    Default is "punct"


normalization: If true, normalizes text into Unicode NFD. 

   Default is false

niemeyer:
Thanks for the updates, @jyemin.

PR #394, opened 2017-02-20
Big merge of fixes from 10gen for 3.4

hexadecy:
- Fix Issue #308: Add Normalization and MaxVariable to Collation struct
- Add collation support for queries.
- Added readPreference and readPreferenceTags to url and DialOptions.
- Fixi issue with getMore not using getMore command when available.


hexadecy:
In travis it lacks test for mongo 3.2, 3.4
But evergreen looks cool

PR #410, opened 2017-03-17
Speed up BSON unmarshalling

craiggwilson:
Currently when unmarshalling into a struct which is missing a field or into bson.Raw, the unmarshaller reads the parts that should be skipped into blackHole. This allows for the entire document to be checked for corruption at the expense of speed. In particular for bson.Raw, it also means that every document is unmarshalled twice, once at the initial time of reading, and later when the bson.Raw is unmarshalled into its final form.  Effectively, this PR stops doing that. It verifies the element that is getting skipped but doesn't descend into them. This is particularly relevent for containers like arrays and documents.

The effect is a massive speedup (I've measured up to 6x) depending on the complexity of documents when using commands that returns cursors as arrays. This would be the new find command and the aggregate command. The downside is that the corruption message appears later in the program than it used to and sometimes a corruption message may not occur if a field is ignored or a bson.Raw is never ultimately unmarshalled. I feel these are acceptable trade-offs.

As part of verifying, I've implemented the entire [bson_corpus](https://github.com/mongodb/specifications/tree/master/source/bson-corpus) as generated code which is checked in (so as long as the corpus doesn't change, no need to regenerate).


fmpwizard:

craiggwilson:
oh, probably need to rebase...

PR #434, opened 2017-05-04
doc fix: Claim what actually happens in the code

unkaktus:


tadukurow (approved):

unkaktus:
ok?

fmpwizard:
If you close and reopen the PR, travis will rerun this build and thanks to  https://github.com/go-mgo/mgo/pull/462 , your PR should pass all tests on all mongodb versions (even though this is just a doc fix that didn't touch any code)

niemeyer:
The documentation is right. The < 0 is an implementation detail and can change, breaking the code if one doesn't follow the documentation.

unkaktus:
Is is not an implementation detail because it defines behavior of the exported function.  What happens if timeout is set to `-2`?

PR #446, opened 2017-05-29
Update session.go add InstantEnsureIndex

sulthonzh:
InstantEnsureIndex

### DIFF
--- a/session.go
+++ b/session.go
@@ -1333,6 +1333,80 @@ NextField:
 	return err
 }
 
+// InstantEnsureIndex is a function for create index from struct schema
+// Example struct { Color   string     `bson:"color" index:"background,sparse"` }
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -1333,6 +1333,80 @@ NextField:
...
+			indexMap[fields] = v
+		} else {
+			indexMap[fields] = []string{key}
+		}
+
+	}
+
+	for k, v := range indexMap {
+		fields := strings.Split(k, "_")
+		index := mgo.Index{}
```
vivace-io, line 1385: This call to package `mgo` is failing the build. Type `Index` already exists in the same package.

vivace-io:
Naming feels rather confusing, doesn't seem to accurately describe what the function does. Might want to consider another name?

hexadecy:
I suggest **EnsureIndexFromStruct**.

PR #462, opened 2017-07-03
Stop all db instances after travis tests

fmpwizard:
If all tests pass, the builds for mongo earlier than 2.6 are still failing.
Running a clean up fixes the issue.


niemeyer:
Thanks!

PR #463, opened 2017-07-04
Set an upper limit of how large we will let txn-queues grow.

jameinel:
When we have broken transaction data in the database (such as from
mongo getting OOM killed), it can cause cascade failure, where that
document ends up getting too many transactions queued up against it.

This can also happen if you have nothing but assert-only transactions against a
single document.

If we have lots of transactions, it becomes harder and harder to add
new entries and clearing out a large queue is O(N^2) which means capping it
is worthwhile. (It also makes the document grow until it hits max-doc-size.)

The upper bound is still quite large, so it should not be triggered if
everything is operating normally.

### DIFF
--- a/txn/flusher.go
+++ b/txn/flusher.go
@@ -212,8 +212,6 @@ var txnFields = bson.D{{"txn-queue", 1}, {"txn-revno", 1}, {"txn-remove", 1}, {"
 
 var errPreReqs = fmt.Errorf("transaction has pre-requisites and force is false")
 
-const maxTxnQueueLength = 1000
-
 // prepare injects t's id onto txn-queue for all affected documents
 // and collects the current txn-queue and txn-revno values during
 // the process. If the prepared txn-queue indicates that there are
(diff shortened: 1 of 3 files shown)


txn/flusher.go
```diff
@@ -244,6 +246,16 @@ NextDoc:
 		change.Upsert = false
 		chaos("")
 		if _, err := cquery.Apply(change, &info); err == nil {
+			if len(info.Queue) > maxTxnQueueLength {
+				// abort with TXN Queue too long, but remove the entry we just added
```
mjs: s/TXN Queue/txn-queue/ ?

txn/flusher.go
```diff
@@ -212,6 +212,8 @@ var txnFields = bson.D{{"txn-queue", 1}, {"txn-revno", 1}, {"txn-remove", 1}, {"
 
 var errPreReqs = fmt.Errorf("transaction has pre-requisites and force is false")
 
+const maxTxnQueueLength = 1000
```
mjs: This seems like a sensible default but it might be good to add support for overriding it on a per session basis in case it's required for some application.

fmpwizard:

rogpeppe:
I don't understand this fully, but my immediate suspicion is that the code looks racy. I'd at least expect a doc comment explaining how/why it's not.

axw (approved):

PR #464, opened 2017-07-04
fix json time zone

reenjii:
Time zone in time format for JSON (un)marshaling is wrong.
All dates used to be parsed in UTC.
See numeric time zone offsets in: https://golang.org/pkg/time/#pkg-constants

I also improved error management when parsing dates so that when an invalid date is encountered, the actual parsing error is returned.

Implemented tests for dates with time zone in test set.


fmpwizard:

Issue #475, opened 2017-08-02
MongoDb simpleIndexKey fails when index is of type int64


mapete94:
The above function for the mongoDB adapter fails when it gets the indexes and the type is int64. The type needs to be accounted for in the if !ok block of that code. I can clean up the code I used for debugging and post it to make it a simple fix.

mapete94:
I opened a PR to fix this issue it's [https://github.com/go-mgo/mgo/pull/476](url)

PR #476, opened 2017-08-02
Fix for Issue #475

mapete94:
[https://github.com/go-mgo/mgo/issues/475](https://github.com/go-mgo/mgo/issues/475])

### DIFF
--- a/session.go
+++ b/session.go
@@ -1537,8 +1537,21 @@ func simpleIndexKey(realKey bson.D) (key []string) {
 		field := realKey[i].Name
 		vi, ok := realKey[i].Value.(int)
 		if !ok {
-			vf, _ := realKey[i].Value.(float64)
-			vi = int(vf)
+			switch realKey[i].Value.(type){
+			case int64:
+				vf, _ := realKey[i].Value.(int64)
+				vi = int(vf)
+			case float64:
+				vf, _ := realKey[i].Value.(float64)
+				vi = int(vf)
+			case string:
+				if vs, ok := realKey[i].Value.(string); ok {
+					key = append(key, "$"+vs+":"+field)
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -1537,8 +1537,21 @@ func simpleIndexKey(realKey bson.D) (key []string) {
 		field := realKey[i].Name
 		vi, ok := realKey[i].Value.(int)
```
niemeyer: This may be included in the type switch as well.

session.go
```diff
@@ -1537,8 +1537,21 @@ func simpleIndexKey(realKey bson.D) (key []string) {
...
+			case float64:
+				vf, _ := realKey[i].Value.(float64)
+				vi = int(vf)
+			case string:
+				if vs, ok := realKey[i].Value.(string); ok {
+					key = append(key, "$"+vs+":"+field)
+					continue
+				}
+			default:
+				panic("Got unknown index key type for field " + field)
```
niemeyer: We pretty much never want to panic on bad data. The driver and the code is not at fault when the data is bogus.

niemeyer (changes requested):
Thanks for the change. Some minor requests below, and a  test for this would be good as well.

mapete94:
I made the requested changes. I did not add a test. I didn't see a way to create a 64bit int for your index type within the documentation. I may have missed it in the documentation, but if there is a way and I'm missing it I can write one.

session.go
mapete94: It's been incorporated into the case statement.

session.go
mapete94: I've removed it, but there is a similar panic right below it that was the reason I included it.

PR #437, opened 2017-05-09
Avoid interrupt inflight requests after a new socket connect failed

carter2000:
If mongoServer fail to dial server, it will close all sockets that are alive, whether they're currently use or not.
There are two cons:
- Inflight requests  will be interrupt rudely.
- All sockets closed at the same time, and likely to dial server at the same time. Any occasional fail in the massive dial requests (high concurrency scenario) will make all sockets closed again, and repeat...(It happened in our production environment)

So I think sockets currently in use should closed after idle.

PR #481, opened 2017-08-17
Avoid interrupt inflight requests after a new socket connect failed

carter2000:
If mongoServer fail to dial server, it will close all sockets that are alive, whether they're currently use or not.
There are two cons:
- Inflight requests  will be interrupt rudely.
- All sockets closed at the same time, and likely to dial server at the same time. Any occasional fail in the massive dial requests (high concurrency scenario) will make all sockets closed again, and repeat...(It happened in our production environment)

So I think sockets currently in use should closed after idle.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -33,9 +33,10 @@ import (
 	"sync"
 	"time"
 
-	"gopkg.in/mgo.v2-unstable/bson"
 	"strconv"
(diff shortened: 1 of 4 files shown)


carter2000:
https://github.com/go-mgo/mgo/pull/437

Issue #489, opened 2017-09-10
bson.ObjectId String() method


Dragomir-Ivanov:
Hi there,
Is there any particular reason for this method to return `ObjectIdHex('abcdef0123456790')` instead of just the hex representation. Since this is not even Mongo type `ObjectID('abcdef0123456790')` I see no reason `String()` function not to return just the plain HEX representation:

```
func (id ObjectId) String() string {
	return id.Hex()
}
```

goofle:
I like to know the reason too.

Rand ahead:
It's funny that this is currently the latest issue. I'm using fatih/structs package to create a `map[string]interface{}` of my structs, and for my current use case your implementation of fmt.Stringer for ObjectId type is getting in the way. I think it would be best if you could remove the method and provide an interface for users to set their own Stringer of ObjectId type, like `func(ObjectId) string`.

goofle:
Although no answers were given to this issue after 9 days, which is very disappointing, I went ahead and made the change and created a pull request at #492.

PR #492, opened 2017-09-19
Exploit SetIdStringer to set id string representation in String method

a-pav:
Although no answers were given to this issue #489, I went ahead made the change I had in mind.


dambrisco:
@goofle Would you be willing to re-apply this to https://github.com/globalsign/mgo ? As this repository's been marked as unmaintained it might gain traction on the new de facto fork.

PR #499, opened 2017-10-11
Add support for allowing partial results from scatter-gather queries

kahuang:
As title indicates, this is useful to allow a scatter-gather query to return partial results when a non-zero number of replica sets are down.


jacksontj:
+1

PR #506, opened 2017-10-17
try to fix socket i/o error for session

zensh:
If socket was killed, it should be removed in session so that new socket can be setup for session.

### DIFF
--- a/session.go
+++ b/session.go
@@ -4408,13 +4408,13 @@ func (s *Session) acquireSocket(slaveOk bool) (*mongoSocket, error) {
 	s.m.RLock()
 	// If there is a slave socket reserved and its use is acceptable, take it as long
 	// as there isn't a master socket which would be preferred by the read preference mode.
-	if s.slaveSocket != nil && s.slaveOk && slaveOk && (s.masterSocket == nil || s.consistency != PrimaryPreferred && s.consistency != Monotonic) {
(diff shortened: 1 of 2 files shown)


zensh:
@niemeyer https://travis-ci.org/go-mgo/mgo/jobs/289072737 this job failed, it should be restart~

zensh:
After this PR，we can use:

```
s := session.Clone()
defer s.Close()
```
instead of 
```
s := session.Copy()
defer s.Close()
```

zensh:
We can try the issue with code:

```go
func main() {
	session, err := mgo.DialWithInfo(someRemoteDailInfo)
	if err != nil {
		panic(err)
	}

	for {
		fmt.Println("Ping:", session.Ping())
		time.Sleep(time.Second)
	}
}
```

Run it, then close the wifi, and then open the wifi, some result like this:
```sh
Ping: <nil>
Ping: <nil>
Ping: <nil>
Ping: <nil>
Ping: <nil>
Ping: <nil>
Ping: <nil>
Ping: read tcp 192.168.0.72:58883->192.168.0.21:27017: i/o timeout
Ping: no reachable servers
Ping: <nil>
Ping: <nil>
Ping: <nil>
```
It means that the session will reconnect automatic.

zensh:
@niemeyer Could you review this PR? We have used the fork version in our productions

PR #518, opened 2017-11-21
Allowed Comment for findAndModify

luby:
findAndModify has a differing syntax for adding comments. We pass the comment through on the Get side, but it just gets lost since the syntax isn't right. One way to fix this is to explicitly pass the Comment through here. I checked the go-mgo/mgo stuff they haven't done this.

This to me is cleaner than trying to parse it out of the bson fields.


luby:
um

